PR #253, opened 2014-08-14
add boot2docker completion

hhatto:
zsh completion for [boot2docker](https://github.com/boot2docker/boot2docker-cli)

### DIFF
--- a/src/_boot2docker
+++ b/src/_boot2docker
@@ -1,0 +1,87 @@
+#compdef boot2docker
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for boot2docker (https://github.com/boot2docker/boot2docker-cli).
+#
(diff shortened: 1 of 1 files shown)


src/_boot2docker
```diff
@@ -0,0 +1,87 @@
...
+
+(( $+functions[_boot2docker_command] )) ||
+_boot2docker_command() {
+  local _boot2docker_cmds
+
+  _boot2docker_cmds=(
+    'init:Create a new boot2docker VM.' \
+    'up:Start VM from any states.' \
+    'start:[Start VM from any states.' \
+    'boot:[Start VM from any states.' \
```
hacfi: Was the leading `[` added on purpose or by accident?

hhatto: @hacfi 
thank you for reporting.

PR #275, opened 2014-12-01
Add completion for the 'mix' Elixir tool

whatyouhide:
Dynamic completion based on the output of `mix help`. Not lightning fast, but works great for me!


whatyouhide:
Closing for inactivity.

PR #278, opened 2015-01-01
Added primusrun script

yforster:
primusrun should replace optirun on the long term. This script is a simple copy of the _optirun version, but with removed arguments.


msva:
Why just don't add it in ~/.zshrc then? It don't use any specific options completion, so, I think, it doesn't need any SPECIAL completion...

PR #290, opened 2015-01-31
Made it a oh-my-zsh plugin and documented how to use it.

ypid:


hacfi:
:+1: cool

hhatto:
nice!! :+1:

caarlos0:
This should also work for antigen, right?

PR #292, opened 2015-02-09
Add completions for `hub` command.

ohcibi:
See github/hub


ohcibi:
Not working. The completions provided from github are not compatible. Never mind.

PR #299, opened 2015-02-27
Update tmuxinator completions

Soliah:
Latest tmuxinator (0.6.10) provides a completions sub command to use for completion purposes. Switch to using that instead of hardcoding.

Related tmuxinator/tmuxinator#262.


hhatto:
Thank you for your contribution.

LGTM

PR #302, opened 2015-03-06
Completion for docker-compose

ilkka:
This change adds not-perfect-but-okay completion for docker-compose.

### DIFF
--- a/src/_docker-compose
+++ b/src/_docker-compose
@@ -1,0 +1,197 @@
+#compdef docker-compose
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for Docker Compose (http://docs.docker.com/compose/).
+#  Adapted from boot2docker completion by hhatto (https://github.com/hhatto)
+#  and docker completion by @aeonazaan and @bobmaerten.
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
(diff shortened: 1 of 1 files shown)


src/_docker-compose
```diff
@@ -0,0 +1,197 @@
...
+  __services
+}
+
+# common args
+_arguments \
+  '--verbose:Show more output' \
+  '--version:Print version and exit' \
+  '--file:Specify an alternate compose file:__yml_files_in_current_dir' \
+  '--project-name:Specify an alternate project name:()' \
+  '*:: :->command'
```
hhatto: You might want to change the following line.

```
_arguments \
  '--verbose[Show more output]' \
  '--version:Print version and exit' \
  '--file[Specify an alternate compose file]:__yml_files_in_current_dir' \
  '--project-name[Specify an alternate project name]:args' \
  '*:: :->command'
```

ilkka:
So finally got around to do this :D thanks for the feedback.

rosstimson:
:+1:  Very useful, especially as `docker-compose` is a lot longer than `fig` and I don't want to have to add yet another alias.

hhatto:
Sorry for late reply.
this patch is merged.

> :+1: Very useful, especially as docker-compose is a lot longer than fig and I don't want to have to add yet another alias.

Please send another pull request :-)

PR #304, opened 2015-03-16
Added rebar completion

rjcoelho:
See https://github.com/rebar/rebar

### DIFF
--- a/src/_rebar
+++ b/src/_rebar
@@ -1,0 +1,79 @@
+#compdef rebar
+
+local curcontext=$curcontext state ret=1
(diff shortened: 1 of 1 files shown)


lrascao:
:+1:

hhatto:
Thank you for contribution.

code is looks good for me.
Could you please add a header?
see [here](https://github.com/zsh-users/zsh-completions/blob/master/src/_ack#L1-L40)

rjcoelho:
Added header with original author and source, but cannot add re-license since code is from oh-my-zsh (MIT). You probably have others similar cases.

PR #311, opened 2015-04-16
Add simple completion function for htop(1).

lucc:


hhatto:
@lucc 
```
compdef _gnu_generic htop
```
Is this no good?

lucc:
yes the completion is nearly identical. But how can I activate the _gnu_generic completion for as may commands as possible? Do I have to maintain a list of commands that support the `--help` option and for which no completion function is provided in my zshrc?

nicoulaj:
You can do this way: https://github.com/nicoulaj/dotfiles/blob/master/.zshrc#L324

PR #312, opened 2015-04-29
Add simple completion function for nload(1).

lucc:
The only thing that is still missing is the completion of network devices as
remaining arguments.  Does anybody know some zsh way to do this or is it best
to parse the output of some command (which one is present in most cases?)?


nicoulaj:
Thanks for your contribution, but I prefer not to merge unfinished work, since people usually just forget about it ;)
Please resubmit it when you feel it is complete.

PR #316, opened 2015-05-24
Add support for local gradle wrapper, gradlew

itegebo:
I'm running zsh 5.0.7 on Mac OS X, installed via brew.  I noticed that while `gradlew` is supported in the compdef, it won't actually execute one's wrapper (assuming current dir is not on path).

### DIFF
--- a/src/_gradle
+++ b/src/_gradle
@@ -40,6 +40,10 @@
 #
 # ------------------------------------------------------------------------------
 
+# Wrapper Support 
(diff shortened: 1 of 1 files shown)


itegebo:
Actually, the current version of _gradle in zsh-completions is not working for me, with or without this patch.

nicoulaj:
Yes, this was unfinished work, and there is big chance it is obsolete now, so I guess we should just remove it,

PR #318, opened 2015-05-26
add github 'hub' tool

jfromaniello:
Add hub completion from here:

https://github.com/github/hub/blob/master/etc/hub.zsh_completion


nicoulaj:
Thanks, but see rules:
https://github.com/zsh-users/zsh-completions#contributing

PR #320, opened 2015-05-27
salt provides its own zsh completion now

gtmanfred:
https://github.com/saltstack/salt/blob/develop/pkg/zsh_completion.zsh

No need for one provided here

https://bugs.archlinux.org/task/44974?project=5&dev=12078

Thanks!
Daniel


hacfi:
:+1:

hhatto:
Thanks.

merged.

PR #321, opened 2015-05-28
add terraform completion

knakayama:


bosr:
+1, should be very useful, thanks!

hhatto:
@knakayama 
Thank you for contribution.

code is looks good for me.
Could you please add a header?
see [here](https://github.com/zsh-users/zsh-completions/blob/master/src/_ack#L1-L40)

hhatto:
Is looks good better to use it
https://github.com/hashicorp/terraform/blob/master/contrib/zsh-completion/_terraform

knakayama:
@hhatto 
Sorry for this PR. I just wanted to add [the terraform completion](https://github.com/hashicorp/terraform/blob/master/contrib/zsh-completion/_terraform) to zsh-completions repository.
I agree with your judgment.

PR #327, opened 2015-06-11
Updating ack & ag.

starcraftman:
* Some opts can repeat.
* Update descriptions.
* Ack was missing some opts present in base 2.00.
* Some cleanup of ag.

~~BEFORE MERGING: I have a question re optspec, for _ag I am trying to allow multiple --ignore opts to be specified. In the manual it suggest just prepend with `*`, this doesn't seem to work in the script. See _ag. Man is a bit sparse on this spec modifier. Do you know why?~~

Someone on IRC tested and seems it is just my local issue. Can whoever pulls verify `ag --ignore dir --ignore` completes the second time --ignore.

### DIFF
--- a/src/_ack
+++ b/src/_ack
@@ -1,4 +1,4 @@
-#compdef ack ack-grep
+#compdef ack ack2 ack-grep ack-standalone
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


hhatto:
LGTM

Thanks.

PR #329, opened 2015-06-18
Add completion for bundle install options

NigoroJr:
Hello,

I added some options for `bundle install` that I found are missing.

The following completions for options are added:
* with
* clean
* full-index
* jobs
* force
* no-cache
* no-prune
* retry
* sheband
* standalone
* trust-policy

See:

https://github.com/bundler/bundler/blob/master/man/bundle-install.ronn

for option details.

Thanks!


NigoroJr:
Thanks for merging!

PR #334, opened 2015-07-18
Update tmuxinator for symlink

hori-ryota:
'find' ignore files in symlink dirs.


Soliah:
This is out of date. Please see #299.

PR #337, opened 2015-07-30
Add completion for gulp

Cronos87:


hhatto:
@Cronos87 
Is there a reason not to use the following files?

https://github.com/gulpjs/gulp/tree/master/completion
https://github.com/gulpjs/gulp/blob/master/completion/zsh

msva:
@Cronos87 
Please, take a look on README.md in the root of the repo, in the ["Contributing" section](https://github.com/zsh-users/zsh-completions#contributing):

PR #339, opened 2015-08-06
Add hub completion from https://github.com/github/hub/blob/master/etc/hub.zsh_completion

jimmidyson:


msva:
Please, take a look on README.md in the root of the repo, in the ["Contributing" section](https://github.com/zsh-users/zsh-completions#contributing):

jimmidyson:
Sorry - should have read it first!

PR #343, opened 2015-08-12
add pidcat

rxwen:
support https://github.com/JakeWharton/pidcat


nicoulaj:
I see this completion is already in the project itself (https://github.com/JakeWharton/pidcat/blob/master/zsh-completion/_pidcat), so I'm closing this.

PR #344, opened 2015-08-13
improve fix for issue-338

rxwen:
- replace all occurence of : to \:
- consider the case there are multiple tcp adb connections
        $ adb devices
        List of devices attached
        192.168.0.139:5555      device
        192.168.0.225:5555      device
  all of colon has to be replaced, otherwise, we get this wrong result:
        $ adb -s 192.168.0.
        192.168.0.139:5555  -- connected device
        192.168.0.225       -- 5555:connected device


msva:
Oh, thanks, I missed it, sorry ;) Merged ;)

PR #346, opened 2015-08-30
Use BSD and Gnu compatible option for tail

nesneros:
Use 'tail -n +2' instead of 'tail +2'. Semanctics are the
same but 'tail +2' is not accepted by Gnu tail.


hacfi:
:+1:

PR #348, opened 2015-09-28
convert ksh functions to POSIX syntax

Eriner:


nicoulaj:
I don't get it, what does this fix exactly ?

PR #351, opened 2015-10-01
Add new commands available in latest django version

mikkeloscar:
This adds missing commands to `manage.py`. Some of the other commands has been deprecated but I have left them to keep support for older django versions.


danechitoaie:
Any news on this? Any idea when it will get merged?

PR #354, opened 2015-10-04
Add a completion script for the primusrun command

gim-:
An extremely small and simple completion script to complete commands from $PATH as an argument for Bumblebee's primusrun command.
(https://github.com/Bumblebee-Project/Bumblebee)


msva:
why don't just `echo compdef _command primusrun optirun >> ~/.zshrc`?

PR #355, opened 2015-10-05
Add complete go 1.5 completion

mikkeloscar:
This adds complete golang 1.5 completions. It is partly based on the completion from oh-my-zsh which should be reflected in the license header.

### DIFF
--- a/src/_go
+++ b/src/_go
@@ -337,7 +337,7 @@ _go() {
                   if (( CURRENT == 2 )); then
                     _arguments \
                       '-func=[output coverage profile information for each function]:string' \
-                      '-html=[generate HTML representation of coverage profile]:string' \
+                      '-html=[generate HTML representation of coverage profile]:file:_files' \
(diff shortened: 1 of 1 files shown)


martinp:
The one in oh-my-zsh was copied from the Go distribution, so there should probably be a "Go authors" copyright in there instead. See this commit for the last revision that existed in the Go repo: https://github.com/golang/go/blob/1ec56062ef6b256f8269e2ca8c5477e3a917331a/misc/zsh/go

It was deleted in https://github.com/golang/go/commit/ddbefae97ef0dd506da22eab811574d57a4f2a3b.

mikkeloscar:
@martinp thanks for letting me know. I have updated authors and copyright year. Since I have based some of it on code made by oh-my-zsh I don't want to leave them out.

martinp:
@mikkeloscar Great. Hoping this will be merged soon.

nono:
Nice! Thanks for adding golang support

PR #356, opened 2015-10-06
Introduced completion for setcap

zv:
This is a custom-written completion for setcap, the default capability manipulation tool in Linux and a widely duplicated standard for capability control on other POSIX systems.


nicoulaj:
Thanks, can you also submit another PR which adds a license header ?

zv:
Sure -- explicit assignment or BSD?

nicoulaj:
You can choose any license you want, see other compdefs.

PR #359, opened 2015-10-20
_docker: Don't cache bogus subcommands list

danielshahaf:
@esc reports that only the `help` subcommand was completed, until he removed `.zcompcache/docker_subcommands`.  I didn't try to reproduce, but here's a patch to hopefully stop this from recurring.


danielshahaf:
Filed upstream at https://github.com/felixr/docker-zsh-completion/pull/27.  Sorry for the noise.

PR #360, opened 2015-10-21
_brew: alfred is no longer a valid cask command

vitorgalvao:
Support for it was removed a good while ago.


joshka:
+1 on merging this. I just ran across this  in zsh and was confused why it was reporting alfred as a cask command.

PR #362, opened 2015-12-16
symfony console completion should also work with php app/console

tunght13488:
Symfony console completation should also work with `php app/console`. Right now it's only work with `php console` or `console`


nicoulaj:
This PR conflicts with another older one, so please resubmit it if you want.

PR #363, opened 2015-12-16
Zsh completion for Hub from Github

tunght13488:
Zsh completion for hub from Github


nicoulaj:
See https://github.com/zsh-users/zsh-completions#contributing

PR #365, opened 2015-12-19
More completion functions, and helpers

vapniks:
This is a dump of a load of completion functions that I have collected and written.
It includes _completion_helpers which contains some new utility functions.


nicoulaj:
Thanks for the contribution, but I will not merge it for these reasons:
 - Please submit one PR per file
 - Each file must have a header (see contributing section in readme)
 - We don't duplicate compdefs here when available in their original project or zsh
 - We want to keep each compdef standalone (no separate file for helper functions), since many people contribute and it's hard to keep it simple

PR #372, opened 2016-02-02
Add git-subtree completion

forivall:
For some reason, completion breaks after the first argument after the subcommand without the `-A 'THIS_IS_A_HACK'`. If anyone could help me with this, that would be great.

For example, without the `-A`, `git subtree add -P subdir <TAB>` doesn't complete.

Otherwise, this is complete.

### DIFF
--- a/src/_git-subtree
+++ b/src/_git-subtree
@@ -1,0 +1,133 @@
+#compdef git-subtree
+#description Merge subtrees together and split repository into subtrees
+------------------------------------------------------------------------------
+# Copyright (c) 2011 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


nicoulaj:
git completion is supported directly in the zsh project (https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_git).

You should submit it to zsh (zsh-workers mailing list).

schlichtanders:
Is this bug already fixed? Or where can I find whether this bug is fixed, and what to do to get the fix?
(as I couldn't find a github issue tracker for zsh-users/zsh, I ask here)

forivall:
@schlichtanders I never got around to submitting this to the zsh mailinglist

Issue #207, opened 2014-01-13
About license


arcan1s:
I have looked at files and found that 50 of 98 files do not have any license. But other 48 files have BSD license and all of them belong to <b>zsh-users</b> (but I could be wrong).
So, I want to suggest:
* To license all of 50 files under BSD.
* To license new files under BSD too and maybe to ask contributors for transfer of copyright to <b>zsh-users</b>. Or maybe to license under BSD new files only.

Why is it needed? We will have a package under one license. And we will not have items with "unknown" license.

[Licenses of 0.11.0 version](https://gist.github.com/arcan1s/8404308)

rrooij:
This is still an important issue in my opinion, since it can't be included in FSF-endorsed distributions because of the missing licenses.

rrooij:
@glidenote 
Can you please add license information to:
* _ag
* _gist
* _httpie
* _rubocop
* _wemux

@sorin-ionescu 
Can you please add license information to:
* _atach
* _gas
* _srm
* _taskwarrior

* _zarg

sorin-ionescu:
@rrooij 
- \_atach: MIT
- \_gas: MIT
- \_srm: MIT
- \_taskwarrior: MIT (missing)
- \_zargs: MIT-like (missing)

mmckinst:
I was just looking at packaging this for Fedora but ran in to this issue.

It looks like some of these are lifted from other sources, assuming there's a copyright there, we can just copy it to the file. For example _brew came from https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/brew_zsh_completion.zsh which has a [BSD license](https://github.com/Homebrew/homebrew/blob/master/LICENSE.txt)

mmckinst:
Down to 52 without a license. https://gist.github.com/mmckinst/3b9049d285c75d3a3572

nicoulaj:
@mmckinst thanks for your work !

To everyone cc'ed, can you please take a look at [mmckinst's table](https://gist.github.com/mmckinst/3b9049d285c75d3a3572) and submit pull requests to add license headers ?

@hhatto @lencioni @dongweiming	@vasilykraev @Valodim @vhbit @shtouff @alexvollmer @tpope @dadrc @benohara @markupboy @technolize @jozefizso @mitukiii @pismute @nono @mafrosis @aki77 @mbrgm

We will wait a bit so that people add license headers to their compdefs, and I guess at some point we will just remove the remaining ones.

[...]

PR #377, opened 2016-02-06
Add license to bower completion

lencioni:
In #207, I was asked to add the BSD license to this completion that I
wrote. This commit takes care of that.


nicoulaj:
Thanks!

Issue #367, opened 2016-01-08
OpenSSL completion


cjeanneret:
Hello,

I'm wondering if it would be possible to get a (working) openssl completion for zsh-5.2…
As I'm working pretty often with this command, having the ability to <TAB> instead of head-scratching and launching a second terminal in order to get openssl sub-commands and options would be a great help :).

I already tested @aschrab openssl branch in his fork, but it doesn't seem to work :/.

Cheers,

C.

black2754:
OpenSSL is a beast. It has a huge number of sub-command, each with their own set of parameters that are not consistent across sub-commands. Documentation from sub-commands (via --help) and the man-pages are often incomplete or missing, contradictory, ambiguous and confusing.
But since I didn't know this back when I thought it a good idea to have completion for openssl, I started out writing something and - a long time later - came up with something halfway usable. I'll make a pull request.

nicoulaj:
Thanks for your work. Since there is remaining work on this, I will leave this issue open so that potential contributors can discuss it here.

PR #384, opened 2016-02-09
cipher suite fix for openssl completion

black2754:
In my openssl completion, I referenced a completion for cipher suites which I did not implement. Sorry for that.
Here it is now, Unfortunately, I could not figure out how to support the +, - and ! prefixes (as in `DEFAULT:!RC4`, for example).


nicoulaj:
#367

PR #392, opened 2016-02-26
Complete cmake property CMAKE_EXPORT_COMPILE_COMMANDS

### LINKED ISSUE #389: Update cmake to include CMAKE_EXPORT_COMPILE_COMMANDS?
Would anyone be opposed to adding [`CMAKE_EXPORT_COMPILE_COMMANDS`](https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) to the [common properties](https://github.com/zsh-users/zsh-completions/blob/6953a4f892df1095e48d43751ccaf9e3ece02190/src/_cmake#L180-L183) that can be completed for cmake? This property has been available since cmake 2.8.5 from mid 2011. This property is used to produce metadata used by many C and C++ development tools that use libclang, for example to add IDE-like features to vim and emacs.

kastiglione:
Adds completion of `-DCMAKE_EXPORT_COMPILE_COMMANDS={YES,NO}`.

Resolves #389


kastiglione:
:tophat:

PR #409, opened 2016-04-07
add docker-compose completion definition

EppO:
Based on https://raw.githubusercontent.com/docker/compose/1.6.2/contrib/completion/zsh/_docker-compose which can be found on the official docker documentation (https://docs.docker.com/compose/completion/)


nicoulaj:
Completions are not accepted when already available in their original project.

PR #410, opened 2016-04-07
Added all docker-machine commands as of 0.6.0

dserodio:
Added `provision`, `regenerate-certs`, `scp`, `status` and `version` docker-machine commands.
`scp` can certainly be improved, but it's a start :)


hsingh23:
looks good to me

PR #415, opened 2016-04-27
Delete _ninja (handled by upstream)

arcan1s:
[Upstream](https://github.com/ninja-build/ninja) already provides own well-maintained completion for some time. Moreover completion is already included to official packages in the latest release (was made at 29 Jun 2015) so it almost has no sense to include ninja completions to this package. (Completion has been added by @netromdk about Apr 4 2016.)


netromdk:
I was unaware of that, sorry. Good point! https://github.com/ninja-build/ninja/blob/master/misc/zsh-completion

PR #417, opened 2016-04-27
enable _adb completion, which is better than zsh's

rxwen:
- try "zsh -s " and tab complete SERIAL, the zsh built-in completion
  function "Lis", which is obviously wrong.
- checked zsh built-in adb completion source code, the issue can be
  walked around by hacking the _wanted dev_serial line
- overall, I consider zsh-completions's _adb has higher quality

### DIFF
--- a/src/_adb
+++ b/src/_adb
@@ -1,0 +1,423 @@
+#compdef adb
+# ------------------------------------------------------------------------------
+# Copyright (c) 2011 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


nicoulaj:
Hi,

`_adb` was removed from this repository because it is also packaged directly in zsh, so we don't want to create confusion.

I suggest you send the better version to zsh-workers mailing list, so that it is included in zsh.

Since I am the original author, I authorize them to do whatever they need with the license/copyright ;)

To be honest I'm not sure which one is the better version, mine must be very outdated now (2011).

PR #423, opened 2016-05-27
add license header to lunchy compdef

markupboy:
re https://github.com/zsh-users/zsh-completions/issues/207#issuecomment-222211724


nicoulaj:
Thank you!

Issue #424, opened 2016-05-29
Copy Docker Completition from official repo


qoomon:
please include following docker completition from official docker repo
https://github.com/docker/docker/blob/master/contrib/completion/zsh/_docker

regards

nicoulaj:
Hi, we will not include docker completion, see our [rules here](https://github.com/zsh-users/zsh-completions#contributing).

PR #425, opened 2016-05-29
feat: add docker completition

qoomon:
official copy of docker completitions from https://github.com/docker/docker/blob/master/contrib/completion/zsh/_docker


nicoulaj:
Closing, see #424.

PR #428, opened 2016-06-14
Add completion for Python Invoke

eligundry:
Copypasta'd from https://raw.githubusercontent.com/pyinvoke/invoke/master/completion/zsh


nicoulaj:
See https://github.com/zsh-users/zsh-completions#contributing

Issue #305, opened 2015-03-30
Homebrew completion file is outdated


number5:
See https://github.com/Homebrew/homebrew/blob/master/Library/Contributions/brew_zsh_completion.zsh

maybe should remove the _brew here just leave a pointer to above file?

nicoulaj:
Thanks, removed.

PR #430, opened 2016-06-27
Add Homebrew Cask completions

joshka:
This completion was originally sourced from the oh-my-zsh brew cask plugin.
(https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/brew-cask/brew-cask.plugin.zsh)
The original plugin suffers from the following issues (IMO):
1. it is part of oh-my-zsh which some have chosen not to use
2. it acts as a script overriding the existing brew completion rather than as a
standard autoloadable file, this means it must be manually installed rather than
automatically installed as part of compinit.

The changes from the plugin are to fix item 2, as well as update the commands to
those current in homebrew cask. This change relies on a change pending in the
homebrew completion to allow it to call subcommand completions (see
https://github.com/joshka/brew/commit/6b010c5700095685bad6797038e83ef558df5104)
and hence should not be merged until that has been merged.

Ideally, this completion would be installed as part of the homebrew cask
installation, but there doesn't seem to be an easy way to do so as it is
homebrew cask is installed as a homebrew tap. It also doesn't really fit in the
homebrew main repo for similar reasons.

See related PRs:
https://github.com/robbyrussell/oh-my-zsh/pull/5191
https://github.com/Homebrew/brew/pull/407

### DIFF
--- a/src/_brew_cask
+++ b/src/_brew_cask
@@ -1,0 +1,83 @@
+#compdef brew-cask
+#autoload
+
+# Autocompletion for homebrew-cask (https://github.com/caskroom/homebrew-cask/).
+#
+# Originally sourced from the oh-my-zsh
+# https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/brew-cask/brew-cask.plugin.zsh
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2009-2016 Robby Russell and contributors
+# See the full list at https://github.com/robbyrussell/oh-my-zsh/contributors
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


pvinis:
why is this and brew not merged?

joshka:
@pvinis I put the 3 PRs up simultaneously about an hour ago, because they're interrelated and understanding one requires the others. Does this answer your question, or have I misunderstood?

pvinis:
oh wow.. i didnt see it was an hour ago.. sorry. im happy to wait for it to be merged. nevermind me then, carry on :)

nicoulaj:
We removed `_brew` in #305 because it was duplicating the one in their repository.
It was followed by a big discussion on their side (Homebrew/legacy-homebrew#49066)

I am not sure I should merge this, after reading [this comment](https://github.com/Homebrew/brew/pull/407#issuecomment-230081558). I agree with @apjanke.

I am closing this PR, please tell me if I misunderstand the situation here.

Issue #436, opened 2016-07-20
Fabric completion raises errors


unklphil:
When autocompleting for fab, I get the following error:

```
_arguments:comparguments:312: invalid argument:
_fab:86: command not found: (--hide=-)--hide=-[comma-separated list of output levels to hide]: :->levels
```

It still does the completion though.

Initially reported here: https://github.com/robbyrussell/oh-my-zsh/issues/5134 but I think this is where it should be reported

nicoulaj:
Solved by #436

PR #437, opened 2016-07-20
Remove trailing spaces

syohex:
This causes #436 issue.
CC: @unklphil


unklphil:
Tried it out, looks like it definitely solves the issue. Thanks!

PR #445, opened 2016-08-08
'pass' added

cbuctok:
A lot of people now using pass, so I decided to add its support to my favourite completions.
Taken from official pass git https://git.zx2c4.com/password-store/plain/src/completion/pass.zsh-completion .

### DIFF
--- a/src/_pass
+++ b/src/_pass
@@ -1,0 +1,144 @@
+#compdef pass
+#autoload
+
+# Copyright (C) 2012 - 2014:
+#    Johan Venant <jvenant@invicem.pro>
+#    Brian Mattern <rephorm@rephorm.com>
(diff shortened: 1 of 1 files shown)


syohex:
[Contributing guideline](https://github.com/zsh-users/zsh-completions#contributing) says

> - Completions are not accepted when already available in their original project.
> - Please do not just copy/paste someone else completion, ask before.

msva:
As already said, since upstream already has a completion, it shouldn't be added here.
It is a place for completions, that upstream refuses to add.

And also we trying to avoid licensing collisions, so taking foreign completions and pushing them here without asking original author for permission (and optionally changing licence, if applicable) — bad idea.

Issue #444, opened 2016-08-05
Most tools in util-linux are missing completions


olejorgenb:
https://en.wikipedia.org/wiki/Util-linux

These have [upstream bash completion definitions](https://git.kernel.org/cgit/utils/util-linux/util-linux.git/tree/bash-completion).

But a quick attempt[1] at using `bashcompinit` didn't work flawlessly.

A crude way of finding out which tools are missing completion definitions:

    # In directory util-linux/bash-completion
    for x in *; do
        if [[ -z $_comps[$x] ]]; then echo $x; fi
    done

This lists almost all the tools. (but it isn't a perfect test. If anyone know of a better way of looking up how a given command is completed I'd be interested)

[1]

    bashcompinit
    source ./lsblk  # is this the correct way?
    lsblk --output F<tab>
    # _lsblk_module:43: command not found: compopt

olejorgenb:
Note: seems https://github.com/RobSis/zsh-completion-generator works pretty well for generating the initial skeleton for.

PR #447, opened 2016-08-09
lsblk completion

olejorgenb:
Completion for lsblk (from util-linux)

I think this should be pretty complete. Only improvement I can see is to add descriptions to the --output colums.

Ref #444


nicoulaj:
Thank you!

PR #448, opened 2016-09-03
adds basic file completion for 'rails test' command

jkraemer:
with rails 5 tests aren't run through rake anymore, but using the `rails test` command.


syohex:
Looks good.

syohex:
And I confirmed it works

syohex:
Thanks

PR #450, opened 2016-09-09
Add _hub for https://github.com/github/hub.

Doron-Behar:
As instructed in "https://github.com/github/hub/tree/master/etc".


syohex:

Please READ https://github.com/zsh-users/zsh-completions#contributing

And you added a symbolic link. All hub users does not install hub into that path.

PR #451, opened 2016-09-26
add --build option to _cmake

pseyfert:
essentially what I posted at skroll/zsh-cmake-completion#1
 * ran RobSis/zsh-completion-generator for minor updates
 * added `--build` and what comes after it:
    * suggest build directory
    * get build targets
    * call `$_comps[make]` to suggest forwarded options to `make` (and similar for `ninja`)
    * suggest e.g. `--clean-first` or `--use-stderr` after `--build`
    * suggest other build options (which were already suggested) at possible points after `--build`
    * check if there has been an "other build option", otherwise keep suggesting the "directly after --build only options"

Some parts basically rewritten, feedback welcome.

### DIFF
--- a/src/_cmake
+++ b/src/_cmake
@@ -1,6 +1,6 @@
 #compdef cmake
 # ------------------------------------------------------------------------------
-# Copyright (c) 2011 Github zsh-users - http://github.com/zsh-users
+# Copyright (c) 2016 Github zsh-users - http://github.com/zsh-users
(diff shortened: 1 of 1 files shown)


syohex:
I got following error at completing with zsh 5.2.

```
_arguments:comparguments:313: invalid argument:
```

BTW please use topic branch instead of master branch for PR. There are some unrelated commits.

PR #452, opened 2016-09-26
add --build option to _cmake

pseyfert:
repetition of #451 with the following changes
 * as suggested not from a master branch
 * fixed `_arguments:comparguments:313: invalid argument:` bug

OP:
essentially what I posted at skroll/zsh-cmake-completion#1

  *  ran RobSis/zsh-completion-generator for minor updates
  *  added `--build` and what comes after it:
       * suggest build directory
       * get build targets
       * call `$_comps[make]` to suggest forwarded options to `make` (and similar for `ninja`)
       * suggest e.g. `--clean-first` or `--use-stderr` after `--build`
       * suggest other build options (which were already suggested) at possible points after --build
       * check if there has been an "other build option", otherwise keep suggesting the "directly after --build only options"
  *  kept the commit by @baberthal in, since it's a fix for _cmake

Some parts basically rewritten, feedback welcome.


syohex:
Looks good.

nicoulaj:
Thanks!

@syohex Feel free to merge if you think everything is OK, you don't need my approval.

syohex:
> @syohex Feel free to merge if you think everything is OK, you don't need my approval.

I understand.

PR #453, opened 2016-10-06
Completion for afew

JindrichPilar:
Completion script for afew an initial tagging script for notmuch
mail.(https://github.com/teythoon/afew)

### DIFF
--- a/src/_afew
+++ b/src/_afew
@@ -1,0 +1,69 @@
+#compdef afew
+# ------------------------------------------------------------------------------
+# Copyright (c) 2016 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


syohex:
Could you put following snippet at end of file ?

```
# Local Variables:
# mode: Shell-Script
# sh-indentation: 2
# indent-tabs-mode: nil
# sh-basic-offset: 2
# End:
# vim: ft=zsh sw=2 ts=2 et
```

And `afew` supports also long option, however this completion supports only short option.

JindrichPilar:
Snippet and long options added. However when you type ``afew -a`` it will disable only adding ``-a`` again, ``--all -n --new`` are still in the posible completions.

PR #407, opened 2016-04-06
WIP: Improved Django completion

blueyed:
This has been in the works since a while locally.

The plan is to get this upstream into Zsh itself, but when using older versions of Zsh etc it might be useful to get into zsh-completions first.

There are some TODOs left, but in general it's good for feedback/testing.

I've merged changes from oh-my-zsh's plugin a while ago (have not checked it now again).

TODO
 - [ ] source attribution(s)

PR #454, opened 2016-10-07
WIP: completion for pytest

blueyed:
This is WIP that needs to be cleaned up.

Inspired/based on https://github.com/alfredodeza/zsh-plugins/blob/master/pytest/pytest.plugin.zsh.

TODO:
 - [ ] caching based on Python sys.path and pytest location, see https://github.com/zsh-users/zsh-completions/pull/407/files#diff-0a6dd68ef4f6d05c6bbafad528caf849R381 and https://github.com/zsh-users/zsh-completions/pull/407/files#diff-0a6dd68ef4f6d05c6bbafad528caf849R453 (because you might have different plugins per venv etc)

### DIFF
--- a/src/_pytest
+++ b/src/_pytest
@@ -1,0 +1,112 @@
+#compdef pytest py.test
+
+local ret=1
+
+local context state state_descr line
+typeset -A opt_args
+
+
+_pytest_caching_policy () {
+  # rebuild if cache is more than a week old
+  local -a oldp
+  oldp=( "$1"(Nm+7) )
+  (( $#oldp ))
+}
+
+local -a pytest_options
+_pytest_options() {
+  # this function calls py.test and loops over the list of apps and options
+  # and constructs the help menu with it
+
+  local cache_policy
+  zstyle -s ":completion:${curcontext}:" cache-policy cache_policy
+  if [[ -z "$cache_policy" ]]; then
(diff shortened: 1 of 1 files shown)


syohex:
Did you read https://github.com/zsh-users/zsh-completions#contributing ?

I suppose you should commit to original repository.

blueyed:
There is no completion shipped with pytest already, but we might (https://github.com/pytest-dev/pytest/issues/1992).
In general I find it more convenient to have it automatically installed by using this repo.
I could also imagine adding it to Zsh itself, but that then needs time to trickle down to users not using the latest.

nicoulaj:
I guess since they don't really have a zsh compdef in their project, we can have it here.

However we don't merge partial work, people tend to just forget about it after they add it here.

blueyed:
Yeah, will reopen then (I've marked it as "WIP" to keep it on the radar, but can see that you rather have the PR queue empty).

For people finding this: I am using this myself, it works good, but needs to get cleaned up for submission here.

blueyed:
@nicoulaj 
Sure, although that contradicts your previous comment.

But I think it is rather complete already in general, and TODOs can be addressed always later.

I've pushed a fixup with my current version (please squash-merge in case you want to take it already).

blueyed:
Hmm, seems to not show up due to the PR being closed?!
The branch is here: https://github.com/blueyed/zsh-completions/tree/pytest

PR #455, opened 2016-10-07
Completion for archlinux-java

JindrichPilar:
Completion script for archlinux-java a tool for selecting default Java runtime (https://wiki.archlinux.org/index.php/java)

### DIFF
--- a/src/_archlinux-java
+++ b/src/_archlinux-java
@@ -41,10 +41,10 @@
 
 _archlinux-java_command_arguments() {
 case $words[1] in
-    (set)
-        local java_versions=("${(@f)$(archlinux-java status | tail -n +2 | tr -s ' ' | cut -d ' ' -f2)}")
-        _describe -t output 'Downloads to delete' java_versions
-        ;;
+  (set)
+    local java_versions=("${(@f)$(archlinux-java status | tail -n +2 | tr -s ' ' | cut -d ' ' -f2)}")
+    _describe -t output 'Downloads to delete' java_versions
+    ;;
 esac
 
 }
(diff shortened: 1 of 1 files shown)


syohex:
Could you add above snippet at end of file and reformat ?

```sh
# Local Variables:
# mode: Shell-Script
# sh-indentation: 2
# indent-tabs-mode: nil
# sh-basic-offset: 2
# End:
# vim: ft=zsh sw=2 ts=2 et
```

JindrichPilar:
Added and reformated

src/_archlinux-java
```diff
@@ -0,0 +1,86 @@
...
+  (set)
+    local java_versions=("${(@f)$(archlinux-java status | tail -n +2 | tr -s ' ' | cut -d ' ' -f2)}")
+    _describe -t output 'Downloads to delete' java_versions
+    ;;
+esac
+
+}
+
+_archlinux-java() {
+local -a commands
```
syohex: Code in function is not indented.

src/_archlinux-java
```diff
@@ -0,0 +1,86 @@
...
+# Authors
+# -------
+#
+#  * Jindřich Pilař (https://github.com/JindrichPilar)
+#
+# ------------------------------------------------------------------------------
+
+
+_archlinux-java_command_arguments() {
+case $words[1] in
```
syohex: Code in function is not indented.

syohex: That snippet is a part of template file of completion file. I don't know well about vim's indentation behaviour.

syohex: Could you set indentation like other completion files ?

Issue #413, opened 2016-04-14
trash-cli


Brottweiler:
Requesting completion for [trash-cli](https://github.com/andreafrancia/trash-cli), like when you tab complete files, the ones you already selected should not show next time you tab complete when deleting multiple files.

PR #456, opened 2016-10-09
Autocompletion for trash-cli

JindrichPilar:
Completion for [trash-cli](https://github.com/andreafrancia/trash-cli)
  * trash-put / trash
  * trash-restore
  * trash-list
  * trash-empty

Implements #413

### DIFF
--- a/src/_trash-empty
+++ b/src/_trash-empty
@@ -41,11 +41,10 @@
 
 _trash-empty() {
 
-
-_arguments -C \
-  '--version[Show programs version number and exit]' \
-  {-h,--help}'[Show help message and exit]' \
-  '*: :' \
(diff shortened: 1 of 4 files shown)


nicoulaj:
Thanks!

Brottweiler:
Are you sure this works with completion of files @JindrichPilar?

I'm using `zstyle ':completion:*:trash:*' ignore-line other` which ignores already completed files;

```sh
t % ls               
four  one  three  two
t % trash one        
 -- file --
four   three  two
```

Without it, only using this completion, it doesn't;

```sh
t % ls                  
four  one  three  two
t % trash one           
 -- file --
four   one    three  two
```

PR #459, opened 2016-10-16
Add completion for envdir

blueyed:
Ref: https://github.com/jezdez/envdir/issues/54

Not sure about the licensing.. I've used the example from the README (https://github.com/zsh-users/zsh-completions/blob/master/src/_ack) as the template.


nicoulaj:
Thanks!

PR #462, opened 2016-10-18
Add completion script for ripgrep

arcizan:


syohex:
Thanks for PR. However I think some options are missing, for example `--case-sensitive`(`-s`) option. Could you check option list again ?

arcizan:
Oops, sorry for my mistake.
I've fixed it.

syohex:
Looks good thanks.

PR #466, opened 2016-11-05
Add completion script for git-journal

niko2342:
add completion script for git-journal a git commit message and changelog generation framework. the completion file was autogenerated by calp-rs.

### DIFF
--- a/src/_git-journal
+++ b/src/_git-journal
@@ -1,0 +1,190 @@
+#compdef git-journal
+
+_git-journal() {
+	typeset -A opt_args
+	local ret=1
+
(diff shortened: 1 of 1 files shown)


syohex:
Please add license header. We cannot merge PR without it.

See also
- https://github.com/zsh-users/zsh-completions#contributing

> Please add a header containing authors, license info, status and origin of the script

Example is here(https://github.com/zsh-users/zsh-completions/blob/master/src/_ack)

niko2342:
Hi @syohex ,

i added the license header, link to the project and names of the authors.

bg
nico

syohex:
Thanks

PR #468, opened 2016-11-17
rg: declare array variables

kastiglione:
Without this, with zsh version `zsh 5.0.8 (x86_64-apple-darwin15.0)`, this error occurs:

    unknown file attribute: \n

For more information see also: https://github.com/rust-lang-nursery/rustup.rs/issues/727

### DIFF
--- a/src/_rg
+++ b/src/_rg
@@ -42,7 +42,8 @@ local context state state_descr line
 local -A opt_args
 local -i ret=1
 
-local common_options=(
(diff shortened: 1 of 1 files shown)


kastiglione:
cc @arcizan

syohex:
Good catch. I have confirmed both zsh 5.2 and 5.08 works with you patch.
Thanks.

PR #470, opened 2016-11-24
yaourt: support .pkg.tar extension

lenoch:
pacman's completion already supports this extension so just follow suit.

Motivation: makepkg produces packages (e.g. from AUR) without compression with its PKGEXT='.pkg.tar' setting, which speeds up building and installing them manually.

### DIFF
--- a/src/_yaourt
+++ b/src/_yaourt
@@ -65,7 +65,7 @@ _yaourt_action_query() {
             _arguments -s : \
                 "$_yaourt_opts_common[@]" \
                 "$_yaourt_opts_query_modifiers[@]" \
-                '*:package file:_files -g "*.pkg.tar.*"'
+                '*:package file:_files -g "*.pkg.tar*"'
             ;;
         query_group)
             _arguments -s : \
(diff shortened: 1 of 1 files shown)


src/_yaourt
```diff
@@ -273,7 +273,7 @@ _yaourt() {
     _yaourt_opts_pkgfile=(
         '-d[Skip dependency checks]'
         '-f[Overwrite conflicting files]'
-        '*:package file:_files -g "*.pkg.tar.*(.)"'
+        '*:package file:_files -g "*.pkg.tar*"'
```
syohex: I suppose it is not necessary to remove `(.)`(matching only files). Why did you remove it ?

lenoch: Thank you for the explanation, I have just pushed a fixed patch.

lenoch:
I force-pushed a fixed patch to the same branch. Thank you for your review :-)

syohex:
Merged

lenoch:
Thanks!

PR #471, opened 2016-12-01
fix version logic for major version > 0

ViktorHaag:
The Silver Searcher ([ag](https://github.com/ggreer/the_silver_searcher)) is now past version 1.0 and so the logic in this completion doesn't cope properly with the fact that while the major version is past 0, the minor version is now less than expected values (testing for values in the 0.20s).

This PR adds in a func to extract the major version and then enhance the comparisons for version-specific option loading to properly account for major version values as well as minor.

This isn't, admittedly, elegant; it's possible that the completion module for `ack` does things better (by exporting version as a `x.y` value and then comparing against that so that 0.25 < 1.0). I'll try using that pattern in this PR to see if I can get it to work, but Zsh scripting is not really in my wheelhouse, so sticking closely to the stuff this completion module was already doing seemed safest.

### DIFF
--- a/src/_ag
+++ b/src/_ag
@@ -40,13 +40,22 @@
 # -*- mode: zsh; sh-indentation: 2; indent-tabs-mode: nil; sh-basic-offset: 2; -*-
 # vim: ft=zsh sw=2 ts=2 et
 # ------------------------------------------------------------------------------
-_ag_version() {
+_ag_minor_version() {
   local version
   version=( $($words[1] --version) )
   version=${version[@]:2:1}
   version=( "${(@s/./)version}" )
   echo "${version[2]}"
 }
+
+_ag_major_version() {
+  local version
+  version=( $($words[1] --version) )
+  version=${version[@]:2:1}
+  version=( "${(@s/./)version}" )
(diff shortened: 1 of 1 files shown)


nicoulaj:
I think the "elegant" way to handle this in zsh completion files is to use `_pick_variant`.
You have an example here: https://github.com/zsh-users/zsh-completions/blob/master/src/_mvn#L81
(In short it will invoke `mvn --version` and set `maven_version=maven3` if the output contains `Maven 3`)

More info here: http://zsh.sourceforge.net/Doc/Release/Completion-System.html

ViktorHaag:
Thanks for the pointer; unfortunately it seems like the variant boundaries here would create a bunch of variants, some of which overlap (i.e. looks like 0.23 added file types and compressed file searching, 0.25 added different heading checking, etc, etc.

I've updated this PR to simply to using the ack completion module's style checking, but stick with the logic structure the original author had here, and that seems safest? If the original author wanted to move to your suggestion, it seems like that might be worthwhile, but a bit complicated to implement, and I kind of think that's their choice to make? I'm just trying to patch their work, not really re-factor too much?

PR #472, opened 2016-12-02
Revert "fix version logic for major version > 0"

ViktorHaag:
Reverts zsh-users/zsh-completions#471

@nicoulaj  -- I thought I had tested the changes in my PR and they were working, but as I attempted to use the changes on another computer I am getting parse errors from the changes in my PR. I don't think it's stable enough to merge, and should get reverted.

Someone else who understands zsh scripting better than me and can do proper testing on completion plugins should probably take this on instead.

I apologize for creating a PR that polluted the main branch. 8(

### DIFF
--- a/src/_ag
+++ b/src/_ag
@@ -40,14 +40,13 @@
 # -*- mode: zsh; sh-indentation: 2; indent-tabs-mode: nil; sh-basic-offset: 2; -*-
 # vim: ft=zsh sw=2 ts=2 et
 # ------------------------------------------------------------------------------
-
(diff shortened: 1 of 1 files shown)


ViktorHaag:
Over the next couple days, I'll see if I can add fixes in for my fixes onto this reversion PR; I won't commit anything here that I can't be really sure works. I'm quite sorry to have bungled this. 8(

nicoulaj:
@ViktorHaag no harm, it's only shell completion ;)

PR #474, opened 2016-12-18
Added raw-json field file path support to httpie-completion.

cjk:
Hi, I noticed httpie-completion was missing support for raw-json file paths so I implemented that in my fork. Please take a look and merge it in if you find it useful.


syohex:
Thanks

PR #486, opened 2017-02-04
added cargo and docker completions

mukimov:


syohex:
Please read https://github.com/zsh-users/zsh-completions#contributing before sending pull request. It says. We cannot accept this PR.

```
- Completions are not accepted when already available in zsh.
- Completions are not accepted when already available in their original project.
- Please do not just copy/paste someone else completion, ask before.
- Completions only partially implemented are not accepted.
- Please add a header containing authors, license info, status and origin of the script (example here).
- Please try to follow Zsh completion style guide.
- Please send one separate pull request per file.
- Send a pull request or ask for committer access.
```

PR #487, opened 2017-02-11
Support to find Ansible hosts file in DebOps style projects

ypid:
In [DebOps](https://debops.org/), it is common to have the Ansible host inventory in a projects directory below "$HOME". This allows to have multiple and separate Ansible environments. For such a setup, the host inventory can be found at `$PWD/ansible/inventory/hosts` as Ansible is typically executed from the root of the project.
This patch adds support for such a layout as fallback.

Ref: https://docs.debops.org/en/latest/debops-playbooks/docs/guides/getting-started.html


syohex:
Merged

PR #488, opened 2017-02-13
Add completion for patool

SmartFinn:
This adds a completion script for the [patool](https://github.com/wummel/patool).


nicoulaj:
Thank you!

PR #490, opened 2017-02-26
Add completion for ffind

SmartFinn:
This adds a completion script for the [ffind](https://github.com/jaimebuelta/ffind).


syohex:
Merged

PR #492, opened 2017-02-28
Add support for completing VBox

marmistrz:
Currently VBoxManage is a symlink to VBox

### DIFF
--- a/src/_virtualbox
+++ b/src/_virtualbox
@@ -1,4 +1,4 @@
-#compdef VBoxManage=vboxmanage VBoxHeadless=vboxheadless
+#compdef VBoxManage=vboxmanage VBoxHeadless=vboxheadless VBox=vboxmanage
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mtorromeo:
On my installation `VBox` can't be used directly.
It is true that the other binaries are links to `VBox` but it works by checking with which name it has been called and if you use only `VBox` it gives an error:
```
VBox --help
Unknown application - VBox
```

syohex:
@mtorromeo Thanks for information. Ah, I checked its sourced. It is shell script and change its behavior by testing `$0` and if `$0` is `VBox` then `VBox` script shows error message and exit.

I closed this PR.

marmistrz:
Ok, but I guess this should be ok: https://github.com/marmistrz/zsh-completions/commit/1869e3808d2910f6e108b9bb591e644b61175e37

Should I open a new pull request? `vboxmanage` is provided by virtualbox on my system.

PR #493, opened 2017-03-03
drop ripgrep completion

arcan1s:
Drop ripgrep completion since the package provides own autogenerated for
each release completion files.


blueyed:
Could we re-add it here, until it's fixed in ripgrep itself?  (see previous comment (on https://github.com/zsh-users/zsh-completions/commit/b7c11f4020246a533cb470d50b94598e190164e9))

kastiglione:
The autogenerated completion is broken (file completion fails) and inferior to this version. I second @blueyed's suggestion to reinstate `_rg`.

PR #496, opened 2017-04-02
Add gist completion for read flag

patrick96:
Fetches and offers completion for the gists of the currently logged in
user (via gist -l) for the '-r' and '--read' arguments
The data is cached for one day before expiring

### DIFF
--- a/src/_gist
+++ b/src/_gist
@@ -75,21 +75,23 @@ _user_gists_cache_policy() {
 user_gists() {
   local update_policy ret=1
   zstyle -s ":completion:${curcontext}:" cache-policy update_policy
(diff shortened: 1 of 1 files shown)


src/_gist
```diff
@@ -61,6 +62,58 @@ _arguments -C \
   '(-P --paste)'{-P,--paste}'[Paste from the clipboard to gist]' \
   '(-h --help)'{-h,--help}'[print options help]' \
   '(-v --version)'{-v,--version}'[print version]' \
+  '(-r --read)'{-r,--read}'[Test]:user gists:user_gists' \
```
syohex: Please fix `[Test]`.

patrick96: Wow. Totally missed that

Will be fixed shortly

PR #497, opened 2017-04-06
Add completions for bitcoin-cli

ianks:
This PR adds completions for `bitcoin-cli`. This is my first completions attempt so any feedback would be awesome! Thanks in advance.


ianks:
@nicoulaj That was what I was hoping. I simply did not have the time to do the entire command set, but figured this was a good start to contribute.

PR #502, opened 2017-05-08
Added tmuxp completions

randrej:
Right now, it only completes ~/.tmuxp projects (xml and yaml), I'll add local file and projectdir completions later.


nicoulaj:
A more complete version of `_tmuxp` has been merged in #536, but thanks anyway!

PR #510, opened 2017-05-26
Restore ripgrep completion

kastiglione:
As mentioned in #493, the (programmatically generated) completion that comes with ripgrep has non-standard completion behavior, in particular no file/path completion. The ripgrep completion contributed to zsh-completions provides more standard/expected completion behavior, and really ought to remain available until the ripgrep bundled completion gets fixed.

### DIFF
--- a/src/_rg
+++ b/src/_rg
@@ -1,0 +1,130 @@
+#compdef rg
+# ------------------------------------------------------------------------------
+# Copyright (c) 2011 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


MaskRay:
Great to find this thread! I have created https://github.com/BurntSushi/ripgrep/pull/496 to move this completion file to the upstream.

nicoulaj:
Against our rules, and it looks like the improved one has now been merged upstream.

kastiglione:
Yep, not necessary. Interesting that inferior/buggy/incomplete completions wouldn't get zsh-completions love, only when absent.

PR #515, opened 2017-07-13
add completion for sublime text

kloetzl:
Added completion for Sublime Text. Hope it is to you liking.

### DIFF
--- a/src/_subl
+++ b/src/_subl
@@ -1,0 +1,60 @@
+#compdef subl
+# ------------------------------------------------------------------------------
+# Copyright (c) 2017 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


src/_subl
```diff
@@ -0,0 +1,60 @@
...
+
+
+_subl() {
+  integer ret=1
+  local I="-h --help -v --version"
+  local -a args
+  args+=(
+    "(- *)"{-h,--help}'[Show help and exit]'
+    "(- *)"{-v,--version}'[Show version and exit]'
+    "($I)--project[Load the given project]:project:"
```
nicoulaj: There is a more elegant way for making `-h --help -v --version` mutually exclusive with all other options; see [_geany](https://github.com/zsh-users/zsh-completions/blob/master/src/_geany) for example

kloetzl: Thanks for the tip, but am I missing something? `--help<TAB>` is still getting completed after another command is already on the command line, which is not what I want.

zsh-completion-huh

nicoulaj: `geany --line 12 --help` is valid, you can try it, it is equivalent to running `geany --help`.

kloetzl: Hm, but then that is not what I want. Even though having `--help` after another option is technically valid, there is no real point in providing a completion. :thinking:

PR #516, opened 2017-07-13
add completion for resilio sync

kloetzl:
This is a completion script for Resilio Sync. I am kind of unsatisfied with the way I handle `ip:port`. But I don't think there is any simple alternative that correctly completes IP addresses.


nicoulaj:
Thanks!

PR #521, opened 2017-08-18
Add completions for ibus

tomonacci:


tomonacci:
@nicoulaj (A belated) thanks (for this and `_xsel`)!

PR #530, opened 2017-09-18
_ibus: avoid the use of positional arguments

tomonacci:
I had a poor understanding of the number of colons in `'*:: :->args'` when I first wrote this. This commit fixes that and as a result lets us remove other hacks in the original code.


tomonacci:
Thanks!

PR #535, opened 2017-10-13
fix #534

### LINKED ISSUE #534: [Yarn] Unable to correct completion if script name contains `:`.
for example, when defined a npm-script as `build:prod`, the completion put out `build -- prod` there.
In the case, `prod` is interpreted as a hints. 
because, ":" is interpreted as a delimiter separating script names and hints.

ktrysmt:
It uses perl-regex.

### DIFF
--- a/src/_yarn
+++ b/src/_yarn
@@ -82,7 +82,7 @@ _yarn_commands_scripts() {
 
 _yarn_scripts() {
   local -a scripts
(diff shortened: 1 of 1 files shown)


ktrysmt:
oh I see, I will try it. thank you.

bddenhartog:
I believe this can be closed, as the original issue this PR addresses (#534) was fixed by https://github.com/zsh-users/zsh-completions/pull/542

PR #536, opened 2017-10-18
Add tmuxp auto-completion

bezhermoso:
* Command auto-completion for `convert`, `freeze`, `import`, and `load`.
* Auto-complete sessions stored in `~/.tmuxp` and current directory.
* Auto-complete for `import tmuxinator` support completion of tmuxinator projects.

### DIFF
--- a/src/_tmuxp
+++ b/src/_tmuxp
@@ -71,7 +71,7 @@ _tmuxp() {
             ;;
           (freeze)
             _arguments -C \
-              '1::session name:compadd $(command tmux ls -F "#{session_name}")'
+              '1::session name:compadd $(command tmux ls -F "#{session_name}" 2>/dev/null)'
             ;;
           (convert)
             _arguments -C \
--- a/src/_tmuxp
+++ b/src/_tmuxp
@@ -47,8 +47,8 @@ _tmuxp() {
   _arguments -C \
     ":command:->command" \
     "*::options:->options" \
-    "--log_level:Log level:(DEBUG INFO WARNING ERROR CRITICAL)" \
-    "--help:Show help"
+    "--log_level:log level:(DEBUG INFO WARNING ERROR CRITICAL)" \
+    "--help[display usage information]"
 
     case $state in
       (command)
@@ -71,7 +71,7 @@ _tmuxp() {
             ;;
           (freeze)
             _arguments -C \
-              '1:: :{_wanted tmux-sessions exp "session name" compadd $(tmux ls -F "#{session_name}")}'
+              '1::session name:compadd $(command tmux ls -F "#{session_name}")'
             ;;
           (convert)
             _arguments -C \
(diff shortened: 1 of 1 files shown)


src/_tmuxp
```diff
@@ -0,0 +1,131 @@
...
+    (sessions)
+      local s
+      _wanted user-sessions expl 'User sessions' compadd \
+        $(
+          for file in ~/.tmuxp/*.(json|yml|yaml)
+          do
+            s=${file:t:r}
+            [[ -z "${line[(r)$s]}" ]] && echo $s
+          done)
+          _wanted sessions expl 'Other sessions' compadd $(for file in ./*.(json|yaml|yml); do echo $file; done)
```
okapia: You're completing both other sessions and user sessions here. If possible use _alternative. I'm also not convinced that the for loops are needed. I don't have tmuxp to test this with but I think this can be reduced to the following:

```
_alternative \
  'sessions-user:user session:compadd -F line - ~/.tmuxp/*.(json|yml|yaml)(:r:t)' \
  'sessions-other:other session:_files -g "*.(json|yaml|yml)(-.)"'
```
Perhaps it needs to be compadd in the second line two; does it allow a path to the file?

src/_tmuxp
```diff
@@ -0,0 +1,131 @@
...
+
+  local curcontext="$curcontext" state line
+  typeset -A opt_args
+
+  _arguments -C \
+    ":command:->command" \
+    "*::options:->options" \
+    "--log_level:Log level:(DEBUG INFO WARNING ERROR CRITICAL)" \
+    "--help:Show help"
+
```
okapia: The "Show help" looks wrong? After the colon, that would be an argument to --help. You probably meant square brackets: `'--help[display usage information]'`

Note also that the zsh convention is to use lowercase for descriptions - "log level".

src/_tmuxp
```diff
@@ -0,0 +1,131 @@
...
+          (load)
+            __tmuxp_load
+            ;;
+          (import)
+            __tmuxp_import
+            ;;
+          (freeze)
+            _arguments -C \
+              '1:: :{_wanted tmux-sessions exp "session name" compadd $(tmux ls -F "#{session_name}")}'
+            ;;
```
okapia: exp here is a variable but you've not declared it local. And in any case, _arguments can provide a description. Does the following work?

`  '1::session name:compadd $(_call_program tmux-sessions ls -F "#{session_name}")'`

bezhermoso: That works, if `tmux-sessions` is replaced with `tmux`. But I seem to be encountering an unrelated issue with `_call_program` not being available. Otherwise that substitution works as intended.

src/_tmuxp
bezhermoso: This worked. New commit is now using `_path_files` as it seem to produce better suggestions.

src/_tmuxp
bezhermoso: Fixed in new commit.

src/_tmuxp
okapia: Sorry, _call_program usage is _call_program _style_ _command_ _args..._
tmux-sessions is the style and the command was apparently 'tmux ls ..." not "ls...".
So both tmux-sessions and tmux should be in there

bezhermoso: @okapia Please see latest commit, made `_call_program` to work using existing usages as guide.

PR #537, opened 2017-10-18
Add tarsnap autocompletion

danteu:


src/_tarsnap
```diff
@@ -0,0 +1,56 @@
...
+#
+# ------------------------------------------------------------------------------
+# Author
+# -------
+#
+#  * Daniel Teunis <daniel@teunis.cc>
+#
+# ------------------------------------------------------------------------------
+
+_tarsnap() {
```
okapia: With zsh autoloadable functions, it is not necessary to wrap the whole file with `_tarsnap() { ... }`. If you do that, you have to add a call at the very bottom: a line saying `_tarsnap "$@"`

As it is, this completion will only work the second time you try to use it in a zsh session: on the first press of <tab>, the real function will be defined but no completions will be added.

I'd suggest you just remove the `_tarsnap() {` bit and drop one level of indentation from the rest of the function.

I'm, aware that many example completion functions appear to do this. It only makes sense if the function defines lots of other functions.

src/_tarsnap
```diff
@@ -0,0 +1,56 @@
...
+    local -a commands
+    commands=(
+        '--fsck:Perform some integrity checks on the archives stored.'
```
okapia: Zsh convention is not to capitalise descriptions and not to use a period/full stop at the end.

src/_tarsnap
```diff
@@ -0,0 +1,56 @@
...
+        '-r:Read the specified archive, convert it to a tar stream, and write it to stdout.'
+        '-t:List archive contents to stdout.'
+        '-x:Extract to disk from the archive.'
+    )
+    
+    if (( CURRENT == 2 )); then
+        _describe -t commands 'commands' commands
+    fi        
+    
+    return 0
```
okapia: Don't return 0. We want to pass on the return status of _describe. Simply removing this line will do the right thing.

src/_tarsnap
```diff
@@ -0,0 +1,56 @@
...
+        '-c:Create an archive containing the specified items and name.'
+        '-d:Delete the specified archive.'
+        '-r:Read the specified archive, convert it to a tar stream, and write it to stdout.'
+        '-t:List archive contents to stdout.'
+        '-x:Extract to disk from the archive.'
+    )
+    
+    if (( CURRENT == 2 )); then
+        _describe -t commands 'commands' commands
+    fi
```
okapia: If CURRENT is not 2, zsh will now complete nothing at all. I'm not familiar with tarsnap but is it really the case that no other further arguments are possible with tarsnap? The golden rule of writing completions is "don't break file completion". If in doubt, call _default from the else branch of this if statement.

PR #543, opened 2017-11-09
Add Mosh completion

shvchk:


okapia:
We don't accept completions that are already available in zsh. Zsh has come with a mosh completion since Jun 2012. That's not to say it can't perhaps be improved, but for that you want to go to the main zsh project.

shvchk:
Ok, thanks, I've changed patch a bit and [sent it to zsh-workers mailing list](https://www.zsh.org/mla/workers/2017/msg01664.html). Closing this.

PR #547, opened 2017-12-23
improve tmuxp load completions

tony:
``'sessions-other:session in current directory:_path_files -g "**/*.(json|yaml|yml)(-.)"'`` can cause a hang with traversing directories and leave the shell inoperable for a few seconds. After the lock up ends, it may either ring the system bell, or return any directory containing a json or yaml file.

This patch limits ``*.(json|yaml|yml)(-.)`` to the current directory to eliminate the hang, while still allowing to complete files. It also uses double splats, but only for finding directories with ``.tmuxp(json/yaml/yml)`` files, which is faster and more accurate. Normal directory completion is also preserved.

1. Limit wildcard json/yaml file matches to current directory
2. Preserve all directory completion (for tabbing into directories)
3. Use unlimited depth search for directories with .tmuxp.(yaml/json)
   inside them. tmuxp load command can load the file or the directory.
4. Preserves behavior of finding files in ``$HOME/.tmuxp``

### DIFF
--- a/src/_tmuxp
+++ b/src/_tmuxp
@@ -101,7 +101,8 @@ __tmuxp_load() {
       local s
       _alternative \
         'sessions-user:user session:compadd -F line - ~/.tmuxp/*.(json|yml|yaml)(:r:t)' \
-        'sessions-other:session in current directory:_path_files -g "**/*.(json|yaml|yml)(-.)"'
+        'sessions-other:session in current directory:_path_files -/ -g "**/.tmuxp.(yml|yaml|json)"' \
+        'sessions-other:session in current directory:_path_files -g "*.(yml|yaml|json)"'
       ;;
   esac
 }


tony:
Any chance this can get looked at, @zsh-users?

The current behavior is broken and causes crashes. This PR fixes it.

Only effects tmuxp completions.

okapia:
I don't think _path_files was ever meant to be used with a recursive glob. What are valid parameters. If you can just specify a session file with a path that needn't necessarily be in the current directory then it should be:

    'sessions-other:other session:_files -g "*.(yml|yaml|json)(-.)"'

If it adds everything from the current directory only, then use compadd directly.

tony:
@okapia Thanks for your reply

When I tried `'sessions-other:other session:_files -g "*.(yml|yaml|json)(-.)"'` it didn't work.

Is there a way I can check logs / debug it?

As it stands, I'm also fine just removing the ``-        'sessions-other:session in current directory:_path_files -g "**/*.(json|yaml|yml)(-.)"'	`` part, since that, at least, is crashing my terminal and at least one other person

okapia:
The usual way to debug these is to press Ctrl+X,?
That puts the full trace of the completion functions in a temporary file. That tends to be rather too much so the & key in less can help with filtering.

I'd also suggest stripping the function and testing just the _alternative part in isolation.

PR #554, opened 2018-02-21
fix "arguments:448: _vim_files: function definition file not found" error

hlecuanda:
I'm not sure if this is the correct place, but it seems as good an upstream as any. 

It's been a fairly recurrent issue to suddenly bump into the error  `_arguments:448: _vim_files: function definition file not found ` as can be seen by googling [`"_vim_files" zsh`](https://www.google.com/search?q="_vim_files"+zsh&oq="_vim_files"+zsh)

I've seen instances of this issue since 2008 (!) and the solution has been to "delete `.zcompdump`" (?) of course this only hides the error until `zcompile` is run again, so i'm submitting this `_vim_files` so hopefully this issue can be put to rest in a way that does not negate the benefits of a compiled completion digest.

Based on the current `_jq` completion it seems to do the job.

### DIFF
--- a/src/_vim_files
+++ b/src/_vim_files
@@ -1,0 +1,61 @@
+#!/usr/bin/env zsh
+# ------------------------------------------------------------------------------
+# Copyright (c) 2011 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the zsh-users nor the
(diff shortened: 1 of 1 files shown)


okapia:
This is not the upstream for either zsh or the vim completion function (which comes with zsh itself). And providing an _vim_files function somewhere in $fpath is not the solution to your problem, it is merely hiding it and you'll probably just get an error for some other function that is defined within a function.

Something has caused your .zcompdump file to be incorrect. Have you got it on an NFS shared directory. You might want to keep separate .zcompdump files by running compinit with, e.g:

    compinit -d ~/.zcompdump.${HOST}.${ZSH_VERSION}

Or somehow you're getting _vim defined without _vim_files. Have you got only the one _vim or is there another one defined in a zcompiled file or from something like oh-my-zsh or antigen. The one that comes with the zsh distribution should be fine. What does the following give you:

    print -l $fpath/_vim(N)

What I would actually suspect is that you are running compinit more than once as part of your zsh startup and the fpath array is getting modified between the two runs. Apart from everything else, this will making starting zsh really slow.

nitrocode:
@okapia that commands returns nothing on my machine. I also responded here https://github.com/robbyrussell/oh-my-zsh/issues/6399

PR #556, opened 2018-03-16
Rename go completion to not collide with zsh file.

### LINKED ISSUE #555: go completion is not loaded
Something screwy is going on with my OMZ env using zsh-completions as a plugin.  Lots of the zsh-completions work with my setup, but it seems that the `_go` file is not read.  Even if I force it with 

`compdef _go go`

the `_go` function it finds is in:

/usr/local/Cellar/zsh/5.4.2_3/share/zsh/functions/_go

Since that is earlier in my `fpath`.

Is anyone else having problems with the golang completion?

jshort:
There is a _go file in the zsh functions directory that would be found
first if you use OMZ with zsh-completions.  This is an ancient file in
zsh and does not complete for the go executable either, but it is still
there on typical zsh distributions.

This change makes it such that zsh/OMZ/zsh-completions can work in
harmony, if one so chooses.

Fixes #555


jshort:
Note that this is just a `git mv` and I removed the wrapping function definition `_go()` and the call to it at the end `_go` since compinit will wrap this in a function named identically to the name of the file (so that it does not collide with _go from the zsh file).

okapia:
Yes, this makes sense given that the _go in zsh completes for different commands such as gccgo.

PR #558, opened 2018-03-18
Add _drill completions

miker:
drill replaced dig
There are no completions for drill on my TrueOS install with the Freebsd
zsh package. The drill command should be a drop in replacement for dig. This
completion file is the _dig completion renamed to work with drill.
Completions for drill work as expected in my testing using this file.

### DIFF
--- a/src/_drill
+++ b/src/_drill
@@ -1,0 +1,106 @@
+#compdef drill
+
+_dns_types() {
+  local expl
+  _description dns-types expl 'DNS type'
+  compadd "$@" "$expl[@]" -M 'm:{a-z}={A-Z}' \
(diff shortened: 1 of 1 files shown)


okapia:
This appears to just be an unmodified copy of the dig completion from the zsh distribution. While dig and drill have some similarities, that aren't that close. Options in particular are all different. You might just do something like compdef _hosts drill. Or try the function here: https://raw.githubusercontent.com/zsh-users/zsh/master/Completion/Unix/Command/_drill

miker:
You are correct. Was just a copy. Thanks for the tip about _drill, a much much better option.

PR #561, opened 2018-04-04
add completion script for Visual Studio Code

Mellbourn:
This script adds command line completion for the `code` command, which is used to start Visual Studio Code (https://code.visualstudio.com/). Code is ranked the most popular developer environment in the [Stack Overflow Developer Survey 2018](https://insights.stackoverflow.com/survey/2018/#development-environments-and-tools).

This implementation gives completion for all documented parameters and has specialized completion for parameters that don't take filenames as parameters, for example `--add` only takes directories as arguments.
I have tested that it works well on macOS and Windows Subsystem for Linux (Ubuntu).

I suggested this script to the maintainers of Visual Studio Code, but they felt it was out of scope of the product itself, and suggested I provide the script externally (https://github.com/Microsoft/vscode/issues/47101). So it would be great if it could be added to zsh-completions instead.

### DIFF
--- a/src/_code
+++ b/src/_code
@@ -1,0 +1,55 @@
+#compdef code
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for Visual Studio Code 1.21 (https://code.visualstudio.com/).
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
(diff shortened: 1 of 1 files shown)


Mellbourn:
I'll close this PR, since the functionality was accepted into `zsh` itself. https://github.com/zsh-users/zsh/pull/24

PR #567, opened 2018-04-27
Add go-task/task completion

sawadashota:
Add completion for `go-task/task`
https://github.com/go-task/task

### DIFF
--- a/src/_task
+++ b/src/_task
@@ -1,0 +1,24 @@
+#compdef task
+
+# Listing commands from Taskfile.yml
+function __list() {
+    local -a scripts
+
+    if [ -f Taskfile.yml ]; then
+        scripts=($(task -l | sed '1d' | sed 's/://' | awk '{ print $2 }'))
+        _describe 'script' scripts
+    fi
+}
+
+_arguments \
+    '(-d --dir)'{-d,--dir}': :_files' \
+    '(-f --force)'{-f,--force} \
+    '(-i --init)'{-i,--init} \
+    '(-l --list)'{-l,--list} \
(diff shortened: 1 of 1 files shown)


okapia:
I'm concerned that the name task conflicts with taskwarrior which is a fairly well-known and widely-used todo-list manager. Taskwarrior includes a zsh completion function named _task too. Including an _task for go-task would likely make zsh-completions conflict with taskwarrior when it comes to operating system packages.

I'd suggest you try to persuade the go-task/task project itself to include the zsh function. At least then anyone packaging taskwarrior and go-task is already dealing with a name clash for task in /usr/bin. Quite a few programs include zsh completions themself: systemd, ripgrep, silver searcher, cargo, curl for example.

As for the function itself, the auxilliary function would be better named _task_scripts than __list. I'm suspicious of the use of _describe for adding matches. Does the array contain a list of values of the form _match_ : _description_. If it is only a list of matches, use compadd directly, perhaps with _wanted, e.g.:
  _wanted scripts expl script compadd -a scripts

Finally, for _arguments, can --dir be --dir=, and for an option named "dir" is _files and not _directories really more appropriate?

You don't need to make the single option --status mutually exclusive with itself. It already will be by default: instead of '(--status)'--status you only need --status. Same applies to --version except, I would suspect that --version can't be used with other options and '(- *)--version' would be appropriate like you have for --help. It'd be nice to have descriptions for the options too.

sawadashota:
@okapia Thank you for your review and suggestion.
I will try persuade the go-task/task project itself to include the zsh function.

PR #568, opened 2018-04-30
glances: update existing arguments

kapsh:


src/_glances
```diff
@@ -40,24 +40,36 @@
 
 _glances() {
```
okapia: I've got a couple of comments to improve this function. They mostly apply to the original function but it would still be good to improve it.

Wrapping zsh autoloadable functions in the function definition is not necessary. That just makes it slower the first time it runs.The file can contain purely the function body.

src/_glances
```diff
@@ -40,24 +40,36 @@
...
-		"-o[Define additional output (available: HTML or CSV)]:output type:(HTML CSV)" \
-		"-p[Define the client or server TCP port (default: 61209)]:port:_ports" \
-		"-P[Client/server password]:password:" \
-		"-s[Run Glances in server mode]" \
-		"-t[Set the refresh time in seconds (default: 3)]:seconds:" \
-		"-v[Display the version and exit]" \
-		"-z[Do not use the bold color attribute]" \
+    _arguments \
+        "-0[Divide task CPU usage by the total number of CPUs]" \
+        "-1[Start Glances in per CPU mode]" \
```
okapia: Zsh convention is to use two spaces for indentation in completion functions and to stick to lowercase for descriptions. So "divide" and "start", however acronyms such as "CPU" can remain in uppercase.

src/_glances
```diff
@@ -40,24 +40,36 @@
...
+        "-B[Bind server to the given IP or host NAME]:host:_hosts" \
+        "-c[Connect to a Glances server]:host:_hosts" \
+        "-C[Path to the configuration file]:configuration path:_files -/" \
+        "-d[Enable debug mode]" \
+        "-h[Display the syntax and exit]" \
+        "-o[Define additional output (available: HTML or CSV)]:output type:(HTML CSV)" \
+        "-p[Define the client or server TCP port (default: 61209)]:port:_ports" \
+        "-q[Disable the curses interface]" \
+        "-s[Run Glances in server mode]" \
+        "-t[Set the refresh time in seconds (default: 3)]:seconds:" \
```
okapia: It is more useful to put hints on units and default values in the heading for the argument so this would be:
`    '-t[set the refresh time]:time (seconds) [3]' \`

src/_glances
```diff
@@ -40,24 +40,36 @@
...
+        "-w[Run Glances in web server mode]" \
+        "-z[Do not use the bold color attribute]" \
+        "--browser[Start the client browser]" \
+        "--disable-bg[Disable background colors in the terminal]" \
+        "--disable-bold[Disable bold mode in the terminal]" \
+        "--hide-kernel-threads[Hide kernel threads in process list]" \
+        "--password[Define a client/server password]" \
+        "--theme-white[Optimize display colors for white background]" \
+        "--tree[Display processes as tree]" \
+        "--username[Define a client/server username]" \
```
okapia: Does the username not need to be specified?

nicoulaj:
@okapia: no answer, but I'll just go ahead and merge it since it is almost only "cosmetic" stuff

PR #570, opened 2018-05-21
adding completion function for chromium (the browser)

pseyfert:
This is quickly written down to help me get the `--proxy-server=socks://` part right.

Can probably be improved (copy and pasted parts of _urls, or letting options exclude each other), but for now this is what I can offer and maybe someone wants to pick it up.


nicoulaj:
Not sure what is the "right" way to do this, let's start from here, thanks

PR #572, opened 2018-05-29
WIP: adding include-what-you-use completion

pseyfert:
This is a rough cut (see inline TODO) but making it public anyway in case someone wants to build on top of it, test for bugs, or just wants to use it as-is.

### DIFF
--- a/src/_include-what-you-use
+++ b/src/_include-what-you-use
@@ -49,7 +49,7 @@ _iwyu_opts() {
              '--cwd=[specify the current working directory]:current working directory:_path_files -/' \
              '--howtodebug[print instructions on how to run iwyu under gdb]' \
              '--howtodebug=[print instructions on how to run iwyu under gdb if file matches argument]:file for debug run:_path_files' \
-             '--mapping_file=[iwyu mapping file]:iwyu mapping file:_path_files -g *.imp' \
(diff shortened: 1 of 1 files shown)


src/_include-what-you-use
```diff
@@ -0,0 +1,65 @@
...
+             "--transitive_includes_only[do not suggest that a file add headers that aren't already visible]" \
+             '--max_line_length=[maximum line length for includes]:a number:' \
+             '--no_comments[do not add "why" comments]' \
+             '--no_fwd_decls[do not use forward declarations]' \
+             '--verbose=[the higher the level, the more output]:a number:'
+}
+
+_arguments "*-Xiwyu[include-what-you-use options]:include-what-you-use options:_iwyu_opts"
+# gcc will also provide --version and --help. Not ideal.
+_gcc
```
nicoulaj: `include-what-you-use` is a superset/wrapper of `gcc` ?

pseyfert: It's usage is `USAGE: include-what-you-use [-Xiwyu --iwyu_opt]... <clang opts> <source file>` so after all the completions from `_arguments "*-Xiwyu...."` there should be compiler completions.

That said, looking at it now it would've been better if I had used `_comps[clang]`.

PR #575, opened 2018-06-11
Improve ansible completion

ssiegel:
Allow stacked/merged and repeated/incremental options


ssiegel:
I just saw there is already PR #527, which will be superseded by this one.

PR #576, opened 2018-06-15
add ccache completion

pseyfert:


pseyfert:
you may have seen that after the initial version I became greedy and made the completion more complex to do even more than just _gnu_generic and forwarding to _gcc. means as of now there are a few FIXMEs left. though still good enough for using.

pseyfert:
From my side I'm done here.

One can probably optimize:
 * `_alternative` for no good reason other than that there I knew the usage w/o looking up a good function
 * avoid code duplication between environment variable and `--set... ` completion

PR #578, opened 2018-06-15
use CPPFLAGS completion for CMAKE_CXX_FLAGS completion

pseyfert:
cmake -DCMAKE_*_FLAGS*=<TAB>
will invoke _gcc as if one completes
export CPPFLAGS=<TAB>

Given that _gcc doesn't distinguish between CFLAGS, CPPFLAGS, or
CXXFLAGS, neither do we here.

### DIFF
--- a/src/_cmake
+++ b/src/_cmake
@@ -35,7 +35,7 @@
 # -------
 #
 #  * Scott M. Kroll <skroll@gmail.com> (initial version)
-#  * Paul Seyfert <pseyfert.mathphys@gmail.com> (handling of --build)
+#  * Paul Seyfert <pseyfert.mathphys@gmail.com> (handling of --build and updates)
 #
 # -------------------------------------------------------------------------
 # Notes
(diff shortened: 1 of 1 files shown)


pseyfert:
Hi,

thinking about it, calling `_gcc` blindly is *probably* usually okay, but I guess it deserves some sanitisation (check if `_gcc` is defined and only then call it, maybe ensure it did something useful with the `-value-` setting … or just forward to the current completion setting for `CPPFLAGS` and don't call `_gcc` specifically / let zsh figure out what to do)

okapia:
I believe the correct way to invoke CPPFLAGS completion is:
`_dispatch cmake -value-,CPPFLAGS,-default-`

Relying on _gcc is not reliable. There's a fair chance of CPPFLAGS being moved to a new _cc in the future where _cc calls either _gcc, _clang, _icc etc.

src/_cmake
```diff
@@ -337,7 +337,7 @@ _cmake_define_property_values() {
     (CMAKE_INSTALL_PREFIX) _files -/ && ret=0;;
     (CMAKE_EXPORT_COMPILE_COMMANDS) _wanted booleans expl 'boolean' _cmake_booleans && ret=0;;
     (CMAKE_*_COMPILER)     _wanted compilers expl 'compiler' _cmake_compilers && ret=0;;
-    (CMAKE_*_FLAGS(|_?*))  _message -e compiler-flags 'compiler flags' && ret=0;;
+    (CMAKE_*_FLAGS(|_?*))  _message -e compiler-flags 'compiler flags' && service=-value-,CPPFLAGS,-default- _gcc && ret=0;;
```
nicoulaj: See @okapia's feedback

pseyfert:
Thanks for looking at this. I tried your suggestion and it works for me. (updated and rebased)

PR #587, opened 2018-07-30
Fix & update hledger completion

fdw:
Unfortunately I had some syntax errors in the last PR. They are now fixed.

Additionally, I added completions for all commands and parameters.


fdw:
I just updated the PR again with a very small improvement. Is there anything else I can do so that it can be merged?

PR #590, opened 2018-08-25
Fix kakoune completion not providing a description argument to _values

daboross:
Using zsh 5.4.2, I occasionally run into an error "_values:compvalues:11: not enough arguments". When it doesn't occur, instead, one of the kakoune sessions is missing from autocompletion.

I believe this is because _values was treating the first session id as a description. As http://zsh.sourceforge.net/Doc/Release/Completion-System.html#Completion-System describes, "The first non-option argument is used as a string to print as a description before listing the values. "

This fixes that by providing a description manually.

### DIFF
--- a/src/_kak
+++ b/src/_kak
@@ -43,7 +43,7 @@
 
 _kak_sessions() {
   session_ids=($(_call_program session_ids kak -l))
-  _values "${session_ids[@]}"
+  _values 'kak sessions' "${session_ids[@]}"
 }
 
 _kak() {
--- a/src/_kak
+++ b/src/_kak
@@ -42,8 +42,10 @@
 # ------------------------------------------------------------------------------
 
 _kak_sessions() {
-  session_ids=($(_call_program session_ids kak -l))
-  _values 'kak sessions' "${session_ids[@]}"
+  local -a session_ids expl
+  session_ids=($(_call_program session_names kak -l))
+  _description session-ids expl "session name"
+  compadd "$expl[@]" -a session_ids
 }
 
 _kak() {


okapia:
Yes, the first argument to _values is a description. However, it seems to me that _values is not the right function for _kak_sessions to be using. _values expects arguments containing descriptions in square brackets and arguments. What does the output of kak -l look like? If a session id could contain certain characters like [ and : then _values is an especially poor choice.

Does the function work if instead of using _values compadd is used directly, for example:
```

  local -a session_ids expl
  session_ids=( $(_call_program session-ids kak -l) )
  _description session-ids expl "session id"
  compadd "$expl[@]" -a session_ids 
```

daboross:
Oh! I'm not entirely sure about what the right function is to use as I'm not a zsh expert. If the description needs to be something different I can do that.

Session IDs can't contain `[` or `:` though.

`kak -l` output looks like:
```console
$ kak -l
4444
tungsten
```

The session names can't contain `[` or `:`, they are limited to ASCII letters, numbers, `_` and `-`. ([code which checks this](https://github.com/mawww/kakoune/blob/2a4d3eb13b325f1633441c6501c1df556b6096ae/src/unicode.hh#L61)).

daboross:
Using `compadd` as you describe works equivalently well. Should I replace my PR with using that?

okapia:
Yes, please change it to use compadd as I described. Even without the concerns of special characters, _values adds a lot of overhead over pure compadd. And note that it was also a bug that the original function didn't declare the session_ids array local as per the first line of my example.

daboross:
Makes sense - thanks!

I have pushed an update to this pr branch.

PR #591, opened 2018-08-25
separate out contribution guidelines and mention upstream/zsh options

okapia:
This moves the contribution guidelines out to CONTRIBUTING.md which is I believe is recognized specially by the github UI and expands on them.

In particular, I wanted to mention the options of contributing directly to zsh or to the upstream project more explicitly and provide some basic information on the advantages each of those approaches might hold.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,39 @@
+## Contributing
+
+### How to Contribute to zsh-completions
+
+Contributions are welcome, just make sure you follow the guidelines:
+
+ * Completions are not accepted when already available in zsh.
+ * Completions are not accepted when already available in their original project.
+ * Please do not just copy/paste someone else's completion, ask before.
+ * Partially implemented completions are not accepted.
+ * Please add a header containing authors, status and origin of the script and license header if you do not wish to use the Zsh license (example [here](src/_ack)).
+ * Any reasonable open source licence is acceptable but note that we recommend the use of the Zsh license and that you should use it if you hope for the function to migrate to zsh itself.
+ * Please try to follow the [Zsh completion style guide](https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide).
+ * Please send one separate pull request per file.
+ * Send a pull request or ask for committer access.
+
+### Contributing Completion Functions to Zsh
+
+The zsh project itself welcomes completion function contributions via
+[github pull requests](https://github.com/zsh-users/zsh/),
+[gitlab merge requests](https://gitlab.com/zsh-org/zsh/) or via patch
+files sent to its mailing list, `zsh-workers@zsh.org`.
+
+Contributing to zsh has the advantage of reaching the most users.
+
+### Including Completion Functions in Upstream Projects
+
+Many upstream projects include zsh completions.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+ * Please try to follow the [Zsh completion style guide](https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide).
+ * Please send one separate pull request per file.
+ * Send a pull request or ask for committer access.
+
+### Contributing Completion Functions to Zsh
+
+The zsh project itself welcomes completion function contributions via
+[github pull requests](https://github.com/zsh-users/zsh/),
+[gitlab merge requests](https://gitlab.com/zsh-org/zsh/) or via patch
+files sent to the mailing list.
```
danielshahaf: Suggest: ``to the mailing list`` → ``to its mailing list, `zsh-workers@zsh.org` ``.

CONTRIBUTING.md
```diff
@@ -0,0 +1,40 @@
...
+function that matches the installed version of their software.
+
+If you are the upstream maintainer this is a good choice. If the project
+already includes completions for bash, fish, tcsh, etc then they are
+likely open to including zsh's too. It can also be a good option for
+completions handling commands that are system or distribution specific.
+
+Ideally, arrange for the project's build system to install the
+completion function in `$prefix/share/zsh/site-functions` where $prefix
+is typically `/usr`.
```
danielshahaf: In this context $prefix should default to `/usr/local`, shouldn't it?

okapia: I've removed the tail end of the note but it might still need some clarification of $prefix. autoconf default is `/usr/local` but people shouldn't hard code that for the install directory.

danielshahaf: > it might still need some clarification of $prefix

How about appending ``where `$prefix` is the installation prefix, such as `/usr/local` or `/opt/<project>` ``?  That should make the meaning clear.

nicoulaj:
> 
It is not, it has to be in a `.github` subdirectory. I just pushed templates for issues/pull requests in a617b745cfca9f4c5f7e4c911dfa3a5bc60c9a3c, feel free to improve on them.

If splitting the docs, is it possible to keep links in the main README ? It is easier for users.

danielshahaf:
@nicoulaj Thanks for adding that!  For future reference, CONTRIBUTING.md may be placed [in a repository's root, `docs`, or `.github` folder](https://help.github.com/articles/setting-guidelines-for-repository-contributors/).

nicoulaj:
My bad, we could have it on root directory then. Does it also apply to issue and PR templates ?

danielshahaf:
> My bad, we could have it on root directory then. Does it also apply to issue and PR templates ?

https://help.github.com/articles/manually-creating-a-single-issue-template-for-your-repository/ (tl;dr: Not sure)
https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/ (tl;dr: Yes)

Personally, I'd put CONTRIBUTING.md in the root directory and leave the others in `.github`: the former is useful for users to run into during casual browsing, but the latter are not.

nicoulaj:
done (and sorry for the mess...)

PR #597, opened 2018-10-03
Add completion script for MacOS Speech Synthesis manager (say)

danteu:


danteu:
Turns out there already exists a completion script for 'say' in the main zsh repo and my local version was simply too old. Closing this PR.

PR #609, opened 2018-11-04
Add _nft (iptables successor) completion

mqus:
Some things to note here:

1. There are bugs in the code as I'm no expert on zsh completion *at all*. I just trial-and-error'd a big part of it.
2. Some things are not done here (marked with a `TODO`), namely the statement syntax of some `nft add` commands.
3. Some completions try to complete existing chain names/rule handles/table names. For this you either have to be root or use sudo AND enable the gain-privileges zstyle (otherwise this will not work)
4. I also found the nft man page to be somewhat lacking so there might be some edge case completions which are plain wrong.

### DIFF
--- a/src/_nft
+++ b/src/_nft
@@ -1,0 +1,465 @@
+#compdef nft
+
+_nft(){
+local -a rules states prev args families options descriptors
+local state="start" line nextstate cmd_obj cmd_subcmd cmd_fam cmd_tab cmd_chain #curcontext="$curcontext"
+
+options=(
+  '(-)'{-h,--help}'[show help]' \
+  '(-)'{-v,--version}'[print version information]' \
+  "(-i --interactive)"{-i,--interactive}'[read input from interactive CLI]: :->end' \
+  "(-f --file)"{-f,--file}'[Read input from <filename>]:nftables rule file:_files' \
+  '(-c --check -n --numeric -N)'{-c,--check}'[check commands validity without actually applying the changes]' \
+  '(-j --json)'{-j,--json}'[format output in json]' \
+  '(-c --check -N)*'{-n,--numeric}'[can be specified up to 3 times, Shows 1:network addresses(default behaviour), 2:Internet services (port numbers) and 3:protocols, user IDs, and group IDs numerically]' \
+  '(-s --stateless)'{-s,--stateless}'[omit stateful information of ruleset]' \
+  '(-N -n --numeric -c --check)'-N'[translate IP addresses to names]' \
+  '(-a --handle)'{-a,--handle}'[output rule handle]' \
+  '(-e --echo)'{-e,--echo}'[echo what has been added, inserted or replaced]' \
+  {-I,--includepath}'[add <directory> to the paths searched for include files. Default is /usr/share]:include directory:_path_files -/' \
+)
+
+# start a state machine. The state is modified by _arguments if the 
+# current argument (descriptors) cannot be completed. Each state has to define is successive state and the 
+# 'descriptors' for _arguments, which essentially tells _arguments how to complete
+_i=0
+while true;do
+  (( _i+=1 ))
+  #Guard for endless loops
+  [[ $_i -gt 100 ]] && return 1
+
+  descriptors=()
+  nextstate="end"
+  case $state in 
+  (start)
+    ##if line is empty (at the start) or ends with semicolon, autocomplete subcommands,
+    # else if we are after a space,complete a semicolon (end the current nft command) and start anew 
+    if [[ $line[1] = "" || $line[1] =~ ';$' ]] ; then
+      descriptors=( ":: :_nft_subcommands" )
+      nextstate="category"
+    else
+      if [[ $words =~ ' $' ]]; then
+        descriptors=(':: :(\;)')
+      else
+        descriptors=(':argument: ')
+      fi
+      nextstate="start"
+    fi
+  ;;
+  (category)
+    case $line[1] in
+    (add | list | flush | delete | create | rename | insert | replace | reset)
+      descriptors=( ":: :_nft_${line[1]}" )
+      nextstate=$line[1]
+    ;;
+    (monitor)
+      descriptors=( ":: : _nft_mon_filter"  )
+      nextstate="mon1"
+    ;;
+    (export)
+      descriptors=( ":: :(ruleset)" ":: :_nft_out_format" )
+      nextstate="preend"
+    ;;
+    (describe)
+      descriptors=( ":expression: ")
+      nextstate="start" #x restart
+    ;;
+    (*)
+      return 1;
+    ;;
+    esac
+
+    #descriptors=( "(ruleset)" )
+    #nextstate="end"
+  ;;
+  (mon1)
+    case $line[1] in
+    (new | destroy)
+#      descriptors=( ":: :_nft_mon_keywords"  ":: :_nft_out_format")
+      descriptors=( ":: : _nft_mon_keywords")
+      nextstate="mon1"
+    ;;
+    (tables | chains | sets | rules | elements | ruleset)
+      descriptors=( ":: : _nft_out_format")
+      nextstate="preend"
+    ;;
+    esac
+  ;;
+  #all completions for create and insert match with the completions of add
+  (create | insert) 
+    state="add"
+  ;|
+  #all completions for reset and flush match with the completions of list
+  (reset | flush)
+    state="list"
+  ;|
+  #(add(^table)/create(^table)/delete/flush(^ruleset)/insert/list(^ruleset)/rename/replace)[family]table
+  (reset | delete | insert | rename | replace | add | create | flush | list)
+    if [[ $state = "add" && $line[1] = "table" ]]; then
+      descriptors=( ":: :_nft_families" ":table name:")
+      nextstate="start" #x restart
+    elif [[ $state = "list" && ( $line[1] = "ruleset" || $line[1] = "tables" ) ]]; then
+      descriptors=( ":: :_nft_families")
+      nextstate="start" #x restart
+    elif [[ $state = "delete" && $line[1] = "table" ]]; then
+      descriptors=(": : _nft_table all-handle")
+      nextstate="tcomplete-delete-table"
+    else
+      cmd_obj=$line[1]
+      cmd_subcmd=$state
+      descriptors=(": : _nft_table all")
+      nextstate="tcomplete"
+    fi
+  ;;
+  (tcomplete-delete-table)
+    # if only a family was completed, complete the table name.
+    case $line[1] in
+    (arp | bridge | inet | ip | ip6 | netdev)
+      descriptors=(": : _nft_table ${line[1]}-handle")
+      cmd_fam=$line[1]
+    ;;
+    # else, complete nothing and go to the next state. default family is 'ip'
+    (*)
+      descriptors=()
+      cmd_fam="ip"
+    ;;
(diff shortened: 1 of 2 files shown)


mqus:
Does it count as 'finished' if it's usable but has some missing subcommands?

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+#  Completion script for nft 0.9.0 (https://www.netfilter.org/projects/nftables/index.html).
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * Markus Richter ( https://github.com/mqus , <mqus@disroot.org>)
+#
+# ------------------------------------------------------------------------------
+_nft(){
```
okapia: Normal convention would actually be to name the file _nftables - i.e. the name of the software rather than the specific command. For example, _smartmontools, _subversion, _perforce etc for smartctl, svn, p4.

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+  '(-N -n --numeric -c --check)'-N'[translate IP addresses to names]' \
+  '(-a --handle)'{-a,--handle}'[output rule handle]' \
+  '(-e --echo)'{-e,--echo}'[echo what has been added, inserted or replaced]' \
+  {-I,--includepath}'[add <directory> to the paths searched for include files. Default is /usr/share]:include directory:_path_files -/' \
+)
+
+# start a state machine. The state is modified by _arguments if the 
+# current argument (descriptors) cannot be completed. Each state has to define is successive state and the 
+# 'descriptors' for _arguments, which essentially tells _arguments how to complete
+_i=0
```
okapia: _i is not declared local anywhere. Underscore prefixes on variables are mostly only used where we intentionally make the variable global and then we use an explicit `typeset -g` to declare them.

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+# ------------------------------------------------------------------------------
+_nft(){
+local -a rules states prev args families options descriptors
+local state="start" line nextstate cmd_obj cmd_subcmd cmd_fam cmd_tab cmd_chain #curcontext="$curcontext"
+
+options=(
+  '(-)'{-h,--help}'[show help]' \
+  '(-)'{-v,--version}'[print version information]' \
+  "(-i --interactive)"{-i,--interactive}'[read input from interactive CLI]: :->end' \
+  "(-f --file)"{-f,--file}'[Read input from <filename>]:nftables rule file:_files' \
```
okapia: Convention is not to capitalise descriptions and references like `<filename>` don't make as much sense in completion descriptions where they aren't listed after the options (as they perhaps were in the man page). I'd suggest something like "read input from specified file".

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+  '(-c --check -n --numeric -N)'{-c,--check}'[check commands validity without actually applying the changes]' \
```
okapia: Isn't there an apostrophe missing in that sentence - "command's".
I think it'd be fine to say "command validity" whereas a missing apostrophe is jarring.
Use double quotes around the description if you want to include apostrophes.

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+  '(-j --json)'{-j,--json}'[format output in json]' \
+  '(-c --check -N)*'{-n,--numeric}'[can be specified up to 3 times, Shows 1:network addresses(default behaviour), 2:Internet services (port numbers) and 3:protocols, user IDs, and group IDs numerically]' \
+  '(-s --stateless)'{-s,--stateless}'[omit stateful information of ruleset]' \
+  '(-N -n --numeric -c --check)'-N'[translate IP addresses to names]' \
+  '(-a --handle)'{-a,--handle}'[output rule handle]' \
+  '(-e --echo)'{-e,--echo}'[echo what has been added, inserted or replaced]' \
+  {-I,--includepath}'[add <directory> to the paths searched for include files. Default is /usr/share]:include directory:_path_files -/' \
```
okapia: As before, use something like "specified directory" rather than "<directory>". Defaults are usually placed in the per-match description in square brackets, e.g.:
```
{-I,--includepath}'[add specified directory to the paths searched for include files]:include directory [/usr/share]:include directory:_directories'
```

Also, does it allow `--includepath=/path` or only `--includepath /path`. If `=` is allowed, include it after the option.

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+	'delete:delete a table, chain, rule, set, element, map, or object'
+	'create:similar to add but returns an error for existing chain'
+	'rename:rename the specified chain'
+	'insert:similar to the add command, but the rule is prepended to the beginning of the chain or before the rule at the given position'
+	'replace:similar to the add command, but replaces the specified rule'
+	'reset:list-and-reset stateful object'
+	'monitor:listen to Netlink events'
+	'describe:show information about the type of an expression and its data type'
+  )
+  _describe -t commands 'nft subcommands' commands "${expl[@]}"
```
okapia: Convention would be for this description should be singular rather than plural "nfs subcommand". The rationale for this being that only one subcommand is typed at this point.

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+}
+_nft_mon_filter(){
+  local filter=(
+    'new:show only events of created objects'
+    'destroy:show only events of deleted objects'
+  )
+  _describe -t filter 'nft monitor' filter -J "action filter" "${expl[@]}"
```
okapia: While tags and groups should be plural (and should probably match) - "filters" or "action-filters". I'd avoid spaces because they're a main to quote in zstyle contexts.

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+  #$1:protocol family
+  #$2:table
+  #$3:obj type (chain/set/map/flowtable/ct helper/counter/quota/meter)
+  #$4:include 'handle'?
+  local objs=( ${(f)"$(_call_program -p nft-$3s nft list table $1 $2 -a 2>/dev/null\
+      | grep ""\\s\*$3"" | sed 's/\s*'"$3"' // ;s/ { # \(.*\)/:(\1)/' )"} )
+  if $4 ;then
+    objs+=( "handle:adress chain by handle")
+  fi
+  _describe -J -t objs objs objs "${expl[@]}"
```
okapia: I'm not to familiar with nftables but is "objs" a meaningful description?

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+}
+
+_nft_obj_handle(){
+  #$1:protocol family
+  #$2:table
+  #$3:obj type (chain/set/ct helper/counter/quota)
+  local handles=( ${(f)"$(_call_program -p nft-$3-handles nft list table $1 $2 -a 2>/dev/null\
+      | grep ""\\s\*$3"" | sed 's/\s*'"$3"' // ;s/ { # handle// ;s/\(\S*\) \(\S*\)/\2:\1/' )"} )
+  _describe -t handles handles handles "${expl[@]}"
```
okapia: use singular for the description - "handle"

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+}
+
+_nft_rule_handle(){
+  #$1:protocol family
+  #$2:table
+  #$3:chain name
+  local rules=( ${(f)"$(_call_program -p nft-rule-handles nft list chain $1 $2 $3 -a 2>/dev/null \
+      |grep -v '^\s*\(table\|chain\|type\|\}\)'|sed 's/^\s*\(.*\) # handle \(\S*\)$/\2:\1/' )"} )
+  _describe -t rules rules rules "${expl[@]}"
```
okapia: ditto.

Also, the tag passed to `_call_program` should ideally match the tag subsequently used when adding the matches.

Are all these calls to `_describe` including descriptions with the matches. I.e., does `rules` here contain _match:description_ _match:description_... or just _match_ _match_…

For just a list of matches, `_describe` is something of a sledgehammer. Instead use compadd direct, perhaps with `_wanted`:
```
_wanted nft-rule-handles expl 'nftables rule handle' compadd "$@" -a rules
```

src/_nft
```diff
@@ -0,0 +1,477 @@
...
+    #remove -handle from $1 to be able to complete table names
+    1="${1/-handle/}"
+  fi
+  case $1 in 
+  (arp | bridge | inet | ip | ip6 | netdev)
+    tables+=( ${(f)"$(_call_program -p nft-tables nft list tables $1 2>/dev/null \
+        | cut -d\  -f3 )"} )
+  ;;
+  esac
+  _describe -V -t tables tables tables "${expl[@]}"
```
okapia: Where did you set the `expl` array? That normally gets setup with `_description` or `_wanted`, but you've not called them in this function.

src/_nft
mqus: should be local, will fix it

src/_nft
mqus: technically nftables has some more commands, e.g. the compat commands like iptables, ip6tables, arptables... but I get the thought behind it

src/_nft
mqus: I removed most capitalizations but must've missed these, I'll fix this

src/_nft
mqus: but there are also completion scripts like `_mvn` (for maven) `_subl` (for sublime text) or `_cf` for Cloud Foundry CLI...
I actually think that _nft is better here.

okapia: > technically nftables has some more commands, e.g. the compat commands like iptables, ip6tables, arptables... but I get the thought behind it

Then we use one function and check `$service` to see which command we're completing for. So e.g. _subversion handles both svn and svn-admin. And yes, the naming isn't a must for the zsh-completions project but it would be renamed if merged to zsh (which is fairly likely given that you opted for the ZSH licence).

src/_nft
mqus: I mean objects (eg. chains, meters, sets,...) which are contained in a 'table'. 
I have no better idea on how to name this... If you do I'll change it.

src/_nft
mqus: I'll use an apostrophe, thanks!

src/_nft
okapia: "objects" would be fine. And better than "objs". At least it's a word from the dictionary. As always, description should be singular, e.g. "object'.

src/_nft
mqus: Yes , I'll fix it

src/_nft
mqus: I've changed these now.

src/_nft
mqus: fixed.

src/_nft
mqus: rules is a list of `match:description`.
In detail it would complete the handle number and show the actual rule associated with this handle number as the description. I've added a comment to explain this.

src/_nft
mqus: done! (description has gotten more specific because of context)

src/_nft
mqus: ...well okay

I'll fix that tomorrow.

src/_nft
nicoulaj: I think these `${expl[@]}` are unnecessary, this should just work:
```suggestion
  _describe -V -t tables tables tables
```

mqus: there are some more `"${expl[@]}"` around, should I remove them all?

mqus: ah right didn't read it completely ;)

mqus:
I'm ready now, are there any other immediate concerns?

nicoulaj:
Thanks again!

PR #618, opened 2019-01-07
Update _rvm

imcvampire:
Update parameter for `ruby`.

`-rubygems` is deprecated 

```'-rubygems.rb' is deprecated, and will be removed on or after 2018-12-01. Remove `-rubygems' from your command-line, or use `-r rubygems' instead  ```


nicoulaj:
The rvm completion is already provided by upstream rvm project, and we have a rule of not duplicating completions here. => deleted `_rvm` on b585ff1

Please see directly with rvm project.

Also, from a quick look I highly doubt your fix is correct, it will either insert `-r\ rubygems` or complete `-r` and `rubygem`.

PR #624, opened 2019-03-11
Add completion for restic

dschulz:


nicoulaj:
Hi, if restic generates its completion we will not duplicate it here. It distro packagers' work to generate the completion and include it in their package.

Issue #397, opened 2016-02-27
Change the gitflow completion to gitflow-avh


minijackson:
As mentioned in [this "thread"](https://github.com/nvie/gitflow/issues/285), the gitflow project is probably dead. Thus, I think moving the completion from gitflow to the gitflow-avh project would be a good idea. Here's the completion proposed in the AUR: https://github.com/petervanderdoes/git-flow-completion/blob/develop/git-flow-completion.zsh (from the same author of the project)

nicoulaj:
If @petervanderdoes wants add his completion to this project, I'll be happy to give him committer access. Otherwise it will stay as it is since it is already maintained in another repo.

petervanderdoes:
I won't mind adding it here, I do have to give a heads up that I'll be unable to do this withing the next few weeks due to a relocation I'm going through. But I'll be more than happy to do this when the dust settles for me.

nicoulaj:
No problem, I gave you committer access.

monkz:
@petervanderdoes may you add your completions?

nicoulaj:
@petervanderdoes never accepted the invitation for committer access. Since he agreed to add the completion, someone else can work on integrating this compdef in the repo (under MIT license).

PR #626, opened 2019-03-12
git-flow: remove non-existing subcommand

yous:
git-flow calls a subcommand based on filename, see [the code of git-flow](https://github.com/nvie/gitflow/blob/15aab26490facf285acef56cb5d61025eacb3a69/git-flow#L91-L97).
But git-flow-status doesn't exist on git-flow repository.


nicoulaj:
I'm not sure what is the status of git-flow, see #397

PR #632, opened 2019-05-18
add options for 'go test -test.'

pseyfert:
Hi,

I noticed `go test` takes a few more arguments that the help does not reveal
and that are not yet provided by the completion. When running `go test -c` to
build a runnable test binary, that binary reveals all these `-test.<something>`
options as `./mytestbinary.test -h`. The arguments are also found to be
accepted by `go test -test.<something> <package>` or `go test <package>
-test.<something>` (i.e. work just like the current `go test` options.

I did not check in which combinations the `-test.` options are meaningful with
other options (e.g. I suspect `go test -c -test.v <package>` is not
meaningful).

Furthermore, I tried around and found that `go test -test.v <PKG1> <PKG2>` and
`go test <PKG1> <PKG2> -test.v` are valid, but `go test <PKG1> -test.v <PKG2>`
isn't.

The completion function is unaware of valid combinations and of the
no-options-between-packages, which is no change to the current behaviour.

As user convenience I bodged the `-test.` addition a bit. I did not want to
spam the entire list of `-test` options on `go test -⇥`, instead I (ab)use the
`-test.-` notation to querry another test function without whitespace after
`-test.`, except the correct completion function is not in my repertoire
(`_arguments` needs leading dashes, `_values` doesn't have per-item
descriptions), so the if branch to a second `_arguments` is likely something a
more versed completion author may want to improve - and I'm curious for
improvement suggestions.

Looking forward to read your comments.

Best,
Paul

### DIFF
--- a/src/_golang
+++ b/src/_golang
@@ -270,34 +270,62 @@ case $state in
           ;;
 
       test)
-        _arguments \
-          "-c[compile but don't run test]" \
-          '-i[install dependencies of the test]' \
-          '-bench[run benchmarks matching the regular expression]:regexp' \
-          '-benchmem[print memory allocation statistics for benchmarks]' \
-          '-benchtime[run benchmarks for t rime]:t' \
-          '-blockprofile[write a goroutine blocking profile to the specified file]:block' \
-          '-blockprofilerate[control goroutine blocking profiles]:n' \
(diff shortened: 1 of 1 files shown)


nicoulaj:
Looks good to me, maybe you could have used `compset -S` or `_parts`, but I guess whatever works.

PR #634, opened 2019-05-31
Completion function for nano editor!

Shivamgupta1234:
- [*] This compdef is not already available in zsh.
- [*] This compdef is not already available in their original project.
- [*] I am the original author, or I have authorization to submit this work.
- [*] This is a finished work.
- [*] It has a header containing authors, status and origin of the script.
- [*] It has a license header or I accept that it will be licensed under the terms of the Zsh license.

### DIFF
--- a/src/_nano
+++ b/src/_nano
@@ -28,11 +28,11 @@
 # Author
 # ------
 # Shivam Gupta <shivam98.tkg@gmail.com>
-# -------------------------------------------------
+# 
 # Description
 # ------------
 # Completion function for nano editor(https://www.nano-editor.org/)!
-# -----------------------------------------------------------
+# 
 
 (( $+functions[_nano_files] )) ||
 _nano_files () {
--- a/src/_nano
+++ b/src/_nano
@@ -1,0 +1,99 @@
+#compdef nano
+# ----------------------------------------------------------
+# Copyright (c) 2019 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


src/_nano
```diff
@@ -0,0 +1,99 @@
...
+# Author
+# ------
+# Shivam Gupta <shivam98.tkg@gmail.com>
+# 
+# Description
+# ------------
+# Completion function for nano editor(https://www.nano-editor.org/)!
+# 
+
+(( $+functions[_nano_files] )) ||
```
nicoulaj: I don't think you need this function, you can directly use `_files` in `_arguments`

src/_nano
```diff
@@ -0,0 +1,99 @@
...
+_nano_files () {
+  case $PREFIX in
+    (*) _files $* ;;
+  esac
+}
+_nano() {
```
nicoulaj: You don't need to wrap in a top level function, you can directly call `_arguments` (eg, like [`_tox`](https://github.com/zsh-users/zsh-completions/blob/master/src/_tox))

src/_nano
```diff
@@ -0,0 +1,99 @@
+#compdef nano
+# ----------------------------------------------------------
+# Copyright (c) 2019 Github zsh-users - http://github.com/zsh-users
```
nicoulaj: If you don't mind, you can cmpletely remove the license header and it will use the Zsh license (this way it could be included directly in Zsh in the future)

Shivamgupta1234: Ok, thanks julien for suggesting change!

nicoulaj:
Thanks for contributing!

PR #636, opened 2019-06-12
provide CMAKE_<LANG>_COMPILER_LAUNCHER in cmake completion

pseyfert:
This variable seems to me worth setting from the commandline.

I'm not sure if "launcher" might not ring a bell in cases "ccache" would. At
least I think for me "e.g. ccache" makes clearer what a launcher is meant to
be.

https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html


nicoulaj:
LGTM, thanks

PR #637, opened 2019-06-13
[cmake] complete CMAKE_GENERATOR environment variable values

pseyfert:
this environment variable will be introduced in cmake 3.15
https://cmake.org/cmake/help/v3.15/manual/cmake-env-variables.7.html#environment-variables-that-control-the-build

At the moment it's a bit bodged in as there is no broad idea for value completion in _cmake.

For now it's okay, I guess, but should there be desire to cover all cmake environment variables that should probably be reworked a bit.

### DIFF
--- a/src/_cmake
+++ b/src/_cmake
@@ -1,4 +1,4 @@
-#compdef cmake
+#compdef cmake -value-,CMAKE_GENERATOR,-default-
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


nicoulaj:
I didn't even know we could this... Thanks!

If you want committer access, you are welcome (you can still send PRs if you want your changes reviewed)

PR #643, opened 2019-07-21
Add completion for inxi

filalex77:
https://github.com/smxi/inxi

- [X] This compdef is not already available in zsh.
- [X] This compdef is not already available in their original project.
- [X] I am the original author, or I have authorization to submit this work.
- [X] This is a finished work.
- [X] It has a header containing authors, status and origin of the script.
- [X] It has a license header or I accept that it will be licensed under the terms of the Zsh license.

### DIFF
--- a/src/_inxi
+++ b/src/_inxi
@@ -1,0 +1,146 @@
+#compdef inxi
+# ------------------------------------------------------------------------------
+# Description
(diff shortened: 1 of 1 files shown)


nicoulaj:
Thanks for contributing!

PR #646, opened 2019-08-23
[golang] prepare for golang 1.13 updates

pseyfert:
In this PR I'm basically following the updates in the release notes of the golang 1.13 RC.

https://tip.golang.org/doc/go1.13#go-env-w

NB: as of opening the PR, 1.13 isn't released yet, I'm just preparing for it. Also, I built this on top of my other PR #642.
The two are independent (642 applies to 1.13 as well as 1.12 - don't know from the top of my head as of when).


nicoulaj:
Thanks!

Do you want committer access by the way ?

PR #650, opened 2019-09-24
Add support for Swift completion

mackoj:
This had been generated with swift 5.1 binary `swift package completion-tool generate-zsh-script > _swift`


nicoulaj:
Thanks, but new compdefs need to tick all the boxes to be accepted.

PR #651, opened 2019-09-26
fall back to file completion for yarn scripts

jordanrastrick:
We generally have a lot of scripts in our various `package.json` that we run via `yarn` (or `yarn run`) and these scripts' arguments are often local files; certainly as the "standard" form of shell completion it seems like the most reasonable fallback option to try when there is no longer anything that can be derived from the `yarn` context.

### DIFF
--- a/src/_yarn
+++ b/src/_yarn
@@ -241,7 +241,7 @@ _yarn() {
         run)
           _arguments \
             '1: :_yarn_scripts' \
-            '*:: :_files'
+            '*:: :_default'
         ;;
 
         tag)
@@ -269,8 +269,7 @@ _yarn() {
         ;;
 
         *) 
-          _arguments \
-            '*:: :_files'
+          _default
         ;;
       esac
     ;;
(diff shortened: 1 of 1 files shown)


src/_yarn
```diff
@@ -266,6 +267,11 @@ _yarn() {
           _arguments \
             '1:query:_files'
         ;;
+
+        *) 
+          _arguments \
+            '*:: :_files'
```
okapia: The correct way to fallback to file completion is to call _default. And just _default, i.e. don't wrap it with _arguments. _arguments does a lot more besides making this rather more inefficient.

In general, such fallbacks are a very good idea. You never know when a new sub-command might be added and breaking file completion is one of the most annoying things a poor programmable completion can do.

jordanrastrick: Thanks for the info, I appreciate it - I am very new to this sort of thing so I really just got the behaviour I was looking for by imitating the rest of the source and verifying it worked in practice, without really fully understanding these sorts of details.

I have amended the PR as suggested.

PR #653, opened 2019-10-02
YARN | Improve suggestions for `yarn run` command

MaximDevoir:
#### New Behavior

> Separate scripts between `package.json` and `node_modules/.bin/`.

zsh-completions-yarn new behavior

#### Old Behavior

> Package scripts and binaries mixed together

oh-my-zsh-yarn-old-behavior

#### Extra

Following [this feedback](https://github.com/robbyrussell/oh-my-zsh/pull/8118#issuecomment-537148142), binaries will always be suggested.

This commit fixes robbyrussell/oh-my-zsh#8115

This commit fixes robbyrussell/oh-my-zsh#8118

### DIFF
--- a/src/_yarn
+++ b/src/_yarn
@@ -81,29 +81,21 @@ _yarn_commands_scripts() {
 }
 
 _yarn_scripts() {
-  local -a commands
-  local -a binaries
-  local -a scripts
-  local -a scriptsUnescaped
-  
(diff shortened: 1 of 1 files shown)


MaximDevoir:
I've noticed that [Node.js JSON](https://stackoverflow.com/a/18469304) could also be used for separating the script from the script command in `package.json`.

Node is probably more present where `yarn` is than `jq`.

PR #660, opened 2019-10-23
Add cppcheck support

jubnzv:


syohex:
Sorry very late reply. I have merged.

PR #661, opened 2019-10-28
Completion for the dbus-launch command

psprint:
I thought that it'll be nice if the command `dbus-launch` would have a completion. So I've created one that I think is of a good quality.

### DIFF
--- a/src/_dbus-launch
+++ b/src/_dbus-launch
@@ -1,0 +1,30 @@
+#compdef dbus-launch
+# Copyright (c) 2019 Sebastian Gniazdowski
+
+setopt localoptions warncreateglobal typesetsilent
+
+local gn ret=1
+
+zstyle -s ':completion:*:*:dbus-launch:*' group-name gn || \
+    zstyle ':completion:*:*:dbus-launch:*' group-name ""
+
+local -a opts
+opts=(
+    --version:"print the version of dbus-launch"
+    --help:"print the help info of dbus-launch"
+    --sh-syntax:"emit bourne-shell compatible code to set up environment variables"
+    --csh-syntax:"emit csh compatible code to set up environment variables"
+    --auto-syntax:"choose --csh-syntax or --sh-syntax based on the SHELL environment variable"
+    --binary-syntax:"use a nul-terminated syntax with the environment data"
+    --close-stderr:"close stderr stream before starting the d-bus daemon"
+    --exit-with-session:"create a persistent \"babysitter\" process"
+    --exit-with-x11:"create a persistent \"babysitter\" process that will connect to the x server"
+    --autolaunch:"scan for a previously-started session and reuse the values found there"
+    --config-file:"pass --config-file=FILENAME to the bus daemon"
+)
+
+_alternative 'files:an optional program to run by dbus-launch:_files' \
+    'opts:option: _describe "dbus-launch options" opts' && \
+        ret=0
+
+return $ret


src/_dbus-launch
```diff
@@ -0,0 +1,31 @@
+#compdef dbus-launch
+# Copyright (c) 2019 Sebastian Gniazdowski
+
+setopt localoptions extendedglob warncreateglobal \
+    typesetsilent noshortloops rcquotes noautopushd
```
okapia: In general, it is best to avoid setting options from completion functions. The contents of $_comp_options is set anyway. Setting rcquotes is especially bad because it can cause completion to insert a match using rcquotes style when a user might not have rcquotes set.

src/_dbus-launch
```diff
@@ -0,0 +1,31 @@
...
+
+local curcontext="$curcontext" gn ret=1
+
+zstyle -s ':completion:*:*:dbus-launch:*' group-name gn || \
+    zstyle ':completion:*:*:dbus-launch:*' group-name ""
```
okapia: I know that the cache system does it but in my own opinion, it is very ugly to be setting styles from a completion function. Styles were designed for user preferences. What are you trying to achieve here?

src/_dbus-launch
```diff
@@ -0,0 +1,31 @@
...
+
+typeset -a opts
+opts=(
+    --version:"Print the version of dbus-launch"
```
okapia: Normal zsh convention is to use lowercase for descriptions, i.e. start this with "print"

src/_dbus-launch
```diff
@@ -0,0 +1,31 @@
+#compdef dbus-launch
+# Copyright (c) 2019 Sebastian Gniazdowski
+
+setopt localoptions extendedglob warncreateglobal \
+    typesetsilent noshortloops rcquotes noautopushd
+
+local curcontext="$curcontext" gn ret=1
```
okapia: Declaring curcontext local is not needed in this function. It is usually only needed when you pass -C to `_arguments` or `_values`.

src/_dbus-launch
```diff
@@ -0,0 +1,31 @@
...
+    --binary-syntax:"Use a nul-terminated syntax with the environment data"
+    --close-stderr:"Close stderr stream before starting the D-Bus daemon"
+    --exit-with-session:"Create a persistent \"babysitter\" process"
+    --exit-with-x11:"Create a persistent \"babysitter\" process that will connect to the X server"
+    --autolaunch:"Scan for a previously-started session and reuse the values found there"
+    --config-file:"Pass --config-file=FILENAME to the bus daemon"
+)
+
+_alternative 'files:an optional program to run by dbus-launch:_files' \
+    'opts:option: _describe "dbus-launch options" opts' && \
```
okapia: Options are normally added with "options" rather than "opts" as the tag.
Any reason for not using `_arguments` here. That would do some things for you. For example, this is not going to honour the prefix-needed style.

Zsh does come with a function named _dbus which handles dbus-send and dbus-monitor. dbus-launch seems to be distributed with those commands. Perhaps you could modify Zsh's _dbus for dbus-launch and contribute it there instead?

src/_dbus-launch
psprint: Yes I've became aware of the completion automatically set options that's why I've removed the `emulate -L zsh`. Just trying to come up with a methodology, i.e.: some standard options that I set in my functions, but if `rcquotes` is harmful then I'll resign out of it. I perhaps leave the `warncreateglobal` and `typesetsilent`.

src/_dbus-launch
psprint: To provide a different default behavior – to separate the `_alternative` options under their descriptions, i.e.: to not look like [here](https://imgur.com/a/E8oE2Gw), where the `file` and `dbus-launch options` are collapsed together.

src/_dbus-launch
psprint: Ok

src/_dbus-launch
psprint: Yes, I've forgot about it, thanks

src/_dbus-launch
psprint: What is the prefix-needed style? I've used `_alternative` because it was what I knew when writing the completion, besides I'm not fully convinced that the options should be completed only when user puts `-` before pressing TAB. AFAIR that's also why I'm using `opts` and not `options`.

Yes I've could do this. I'm doing the PRs to accomplish Hacktoberfest, but the completion-way of doing this is so easy and meaningful, and there are 3 days left, that I can resign of the dbus-launch completion and submit it to the Zsh upstream.

src/_dbus-launch
okapia: You may not like zsh's default grouping behaviour (neither do I much but there are reasons for it) but changing it per-completion is just confusing for anyone configuring things. It really is best left alone in individual functions.

src/_dbus-launch
okapia: If you want options completed without the initial `-`, then configure the prefix-needed style in your .zshrc. I have it configured specificially for certain styles.

When writing completion functions, just adhere to the defaults and check common styles such as prefix-needed. Or use helpers like _arguments which save you the work.

psprint: I didn't know about the prefix-needed style. I can see the way of expecting user to grasp the completion at some point of using Zsh and to configure the styles as needed. On the other hand, if a different prefix behavior has sense for a particular command, would anyone bother to change it, if it's such a niche command like `dbus-launch`? I can see the effort happening for some more commonly used commands, but maybe not for such a niche commands. That said, I don't think that `dbus-launch` is a command that specifically gains from the no-prefix behavior, so I think that I change it to use _arguments.

Yes, it would work – knowing that DOcean cannot / will not verify the PRs, but would the PR be merged? I guess it would be just symbolical? On the other hand, it would be justified and fine – I would be really contributing to an OS project.

src/_dbus-launch
psprint: As I really don't like the default grouping behavior, I would prefer to keep the style.  Unless, of course, I'll be contributing to Zsh upstream.

psprint:
@okapia: I've updated the dbus completion and created the PR: https://github.com/zsh-users/zsh/pull/41

PR #663, opened 2019-10-28
Completion for the Gentoo `eix' command

psprint:
I've found that many of Gentoo tools don't have a Zsh completion. Here's one for the useful `eix` command.


psprint:
(Un)fortunately, the command has its completion at its base project, closing the PR

PR #664, opened 2019-10-28
Completion for the `aplay' command (from alsa-utils)

psprint:
Spotted one other useful command that doesn't have a completion – the `alsa-utils` command `aplay`. The PR provides the completion.

### DIFF
--- a/src/_alsa-utils
+++ b/src/_alsa-utils
@@ -1,0 +1,54 @@
+#compdef aplay
+# Copyright (c) 2019 Sebastian Gniazdowski
+
+setopt localoptions warncreateglobal typesetsilent
+
+local -a opts
+opts=(
+  + '(operation)'
+  {-h,--help}'[print help message]'
+  --version'[print current version]'
+  {-l,--list-devices}'[list all soundcards and digital audio devices]'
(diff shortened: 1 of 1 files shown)


psprint:
> The function should arguably be named _alsa-utils. Zsh similarly, has _subversion and _smartmontools rather than _svn, _smartctl. It might then later pick up support for alsamixer, alsactl etc.

I've renamed the file.
 
> The ret variable is not needed, a shell function will pass on the return status from the last command it executes. So if _arguments is the last command, the return status gets passed on. Similarly, you don't need the local curcontext (or to pass -C to _arguments). This is only needed if you use the _arguments' states feature.

I've made the two changes.
 
> The point I made about setting local options on one of your other PRs also applies here.

Done.
 
> As a potential improvement, consider adding some exclusion lists to the _arguments specs.

I've added:

```
  …
  + '(operation)'
  {-h,--help}'[print help message]'
  --version'[print current version]'
  {-l,--list-devices}'[list all soundcards and digital audio devices]'
  {-L,--list-pcms}'[list device names]'
  + option
  {-D,--device}'[select PCM by name]'
  …
```
```

I'll now create a PR to the zsh-users/zsh repo.

PR #666, opened 2019-11-02
Yarn | cross-platform sed replacement pattern newline solution

### LINKED ISSUE #662: Autocomplete bug with yarn run
Autocomplete for yarn looks like this:

image

I checked a few commits. Seems that bug was introduced in https://github.com/zsh-users/zsh-completions/commit/a47b58ef97ad489f19fc1ecb9933597eef9a927a#diff-7d8d4a411c6bd5f832a76535275f89e1

Is it related to the line-break symbol?

MaximDevoir:
Some sed utilities (e.g. Mac OS / BSD) don't interpret `\n` in a replacement pattern as a newline. See the discussion [here](https://www.superuser.com/q/307165).

Solution tested at https://github.com/MaximDevoir/zsh-completions-662/runs/286153577 on Mac and Ubuntu

Fixes #662

### DIFF
--- a/src/_yarn
+++ b/src/_yarn
@@ -86,9 +86,11 @@ _yarn_scripts() {
   local i runJSON
 
   runJSON=$(yarn run --json 2>/dev/null)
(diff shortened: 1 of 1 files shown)


MaximDevoir:
I'm going to go ahead and merge this PR since there haven't been any objections from other members on the team.

I've been using this PR along with #668 in my local oh-my-zsh copy of `_yarn` for the past few months without encountering any errors.

PR #675, opened 2019-12-22
Add completion for fail2ban-client

xdch47:
- [x] I am the original author
- [x] I accept that it will be licensed under the terms of the Zsh license.

### DIFF
--- a/src/_fail2ban-client
+++ b/src/_fail2ban-client
@@ -1,0 +1,312 @@
+#compdef fail2ban-client
+
+# ------------------------------------------------------------------------------
+# Description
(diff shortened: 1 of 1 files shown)


xdch47:
cross-platform replacement pattern newline solution (likewise #666).

syohex:
Sorry very late reply. macOS `sed` command does not support `-r` option.  Could you rewrite `sed` code without `-r` option ?

xdch47:
okay - escaped `+` and `(` `)` in sed regex's

syohex:
Thanks

PR #683, opened 2020-03-04
update golang for go 1.14 and improve go env handling

pseyfert:
standard PR header:

- [n/a] This compdef is not already available in zsh.
- [n/a] This compdef is not already available in their original project.
- [✓] I am the original author, or I have authorization to submit this work.
- [✓] This is a finished work.
- [✓] It has a header containing authors, status and origin of the script.
- [✓] It has a license header or I accept that it will be licensed under the terms of the Zsh license.

what I did is basically:

 - read the changelog of go 1.14 and updated _golang accordingly
 - noticed the `go env` completion was written a bit cumbersome (probably by me), cleaned that up a bit
 - added expansion for go environment variables after `go env <SOMEVAR>=<TAB>` and for "regular" shell variable expansion `GO<SOMETHING>=<TAB>`
 - added completions for `go build -gcflags=<TAB>` an d`go build -gcflags=-d=<TAB>`

### DIFF
--- a/src/_golang
+++ b/src/_golang
@@ -43,8 +43,6 @@
 #
 # ------------------------------------------------------------------------------
 
-typeset -A opt_args
-
 __go_buildmodes() {
   local -a buildmodes
   buildmodes=(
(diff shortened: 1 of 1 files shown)


syohex:
Thanks for PR. I have applied your commits.

PR #685, opened 2020-03-10
Add eksctl completion

ilpianista:
Add completion for [eksctl](https://github.com/weaveworks/eksctl).

Generated with `eksctl completion zsh` (which uses [cobra](https://github.com/spf13/cobra) under the hood to generate it).


syohex:
> This compdef is not already available in their original project.

This is available in original project by `eksctl completion zsh`. I suppose it is not good to add this compdef to `zsh-completions` because we need to update the compdef when original project is updated. It makes maintenance hard.

Issue #633, opened 2019-05-21
node --prof-process flag doesn't autocomplete *.log files


reasonablytall:
The node autocompletion restricts to *.js files which works well for the general use case, but when processing a profile log with `--prof-process` it becomes difficult to complete the *.log files generated by `--prof`.

I'm not sure if the completion system can handle it, but it would be nice if it would switch to completing *.log files when the `--prof-process` flag is passed.

syohex:
I have sent PR #690

syohex:
I've fixed this issue. Please check latest version.

PR #690, opened 2020-03-21
Complete only .log file if --prof-process is specified

syohex:
Fixes: https://github.com/zsh-users/zsh-completions/issues/633


syohex:
Thanks for review.

Issue #677, opened 2020-01-20
Completion for `kitchen` doesn't list all options


msiebuhr:
`kitchen`, used for setting up quick test-servers, has this help-overview (which is what zsh-completions uses to parse out commands and help):

```sh
$ kitchen help
Commands:
  kitchen console                                 # Kitchen Console!
  kitchen converge [INSTANCE|REGEXP|all]          # Change instance state to converge. Use a provisioner...
  kitchen create [INSTANCE|REGEXP|all]            # Change instance state to create. Start one or more i...
  kitchen destroy [INSTANCE|REGEXP|all]           # Change instance state to destroy. Delete all informa...
  kitchen diagnose [INSTANCE|REGEXP|all]          # Show computed diagnostic configuration
  kitchen doctor INSTANCE|REGEXP                  # Check for common system problems
  kitchen exec INSTANCE|REGEXP -c REMOTE_COMMAND  # Execute command on one or more instance
  kitchen help [COMMAND]                          # Describe available commands or one specific command
  kitchen init                                    # Adds some configuration to your cookbook so Kitchen ...
  kitchen list [INSTANCE|REGEXP|all]              # Lists one or more instances
  kitchen login INSTANCE|REGEXP                   # Log in to one instance
  kitchen package INSTANCE|REGEXP                 # package an instance
  kitchen setup [INSTANCE|REGEXP|all]             # Change instance state to setup. Prepare to run autom...
  kitchen test [INSTANCE|REGEXP|all]              # Test (destroy, create, converge, setup, verify and d...
  kitchen verify [INSTANCE|REGEXP|all]            # Change instance state to verify. Run automated tests...
  kitchen version                                 # Print Kitchen's version information
```

But doing tab-completion for the commands, **login**, **doctor**, **exec** and **package** is missing:

```sh
$ kitchen <tab><tab>
console   -- Kitchen Console!
converge  -- Change instance state to converge. Use a provisioner...
create    -- Change instance state to create. Start one or more i...
destroy   -- Change instance state to destroy. Delete all informa...
diagnose  -- Show computed diagnostic configuration
help      -- Describe available commands or one specific command
init      -- Adds some configuration to your cookbook so Kitchen ...
list      -- Lists one or more instances
setup     -- Change instance state to setup. Prepare to run autom...
test      -- Test (destroy, create, converge, setup, verify and d...
verify    -- Change instance state to verify. Run automated tests...
version   -- Print Kitchen's version information
```
[...]

PR #692, opened 2020-03-23
kitchen: fix regexp which collects subcommands

syohex:
fix: https://github.com/zsh-users/zsh-completions/issues/677

##### before

before

##### apply this patch

after


syohex:
Thanks for review.

Issue #41, opened 2011-08-11
Android 'emulator' completion


nicoulaj:
Add completion for Android `emulator` command.

PR #696, opened 2020-03-27
Add avd completion to _emulator

nfischer:
This resolves a TODO in _emulator by adding completion for AVD names.
This adds completion for "-avd" as well as the special "@AVD_NAME"
syntax for non-option arguments.

Issue #41

### DIFF
--- a/src/_emulator
+++ b/src/_emulator
@@ -18,11 +18,15 @@
 typeset -A opt_args
 local context state line curcontext="$curcontext"
 
+_list_avds() {
+  local -a _avds=($HOME/.android/avd/*.ini(N.:t:r))
+  echo "${_avds[@]}"
+}
+
 # TODO All image options are contextual to -sysdir value
 # TODO All skin options are contextual to -skindir value
 # TODO snapshot options are mutually exclusive
 # TODO Use '-snapshot-list' output for snapshot names
-# TODO Complete -avd with avd names ?
 # TODO -logcat: use completer from _adb
 # TODO Complete options with device values
 # TODO Complete -prop
@@ -62,7 +66,7 @@ _arguments \
   '-snapshot-list[show a list of available snapshots]' \
   '-no-snapshot-update-time[do not do try to correct snapshot time on restore]' \
   '-wipe-data[reset the user data image (copy it from initdata)]' \
-  '-avd[use a specific android virtual device]:android virtual device name' \
+  '-avd[use a specific android virtual device]:android virtual device name:($(_list_avds))' \
   '-skindir[search skins in <dir> (default <system>/skins)]: :_files -/' \
   '-skin[select a given skin]' \
   '(-noskin -no-skin)'{-noskin,-no-skin}'[don'\''t use any emulator skin]' \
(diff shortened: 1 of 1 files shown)


src/_emulator
```diff
@@ -18,11 +18,18 @@
 typeset -A opt_args
 local context state line curcontext="$curcontext"
 
+_list_avds() {
+  local -a _avds=("${(f)$(cd $HOME/.android/avd/ && ls *.ini)}")
```
nfischer: As far as I can tell, "$HOME/.android/avd/" is a hardcoded path in the emulator binary. I could replace this with `emulator -list-avds`, but many folks don't ever add `emulator` to their $PATH, so I thought this might work more universally.

src/_emulator
```diff
@@ -108,7 +115,21 @@ _arguments \
   '-memcheck[enable memory access checking]:flags' \
   '-qemu[pass arguments to qemu]:arguments' \
   '-verbose[same as '\''-debug-init'\'']' \
-  '*'{-debug,-debug-,-debug-no-}'[enable/disable specific debug messages]:tag'
+  '*'{-debug,-debug-,-debug-no-}'[enable/disable specific debug messages]:tag' \
+  '1: :->cmds' \
+  '*:: :->args' && ret=0
+
+case $state in
+  cmds)
```
nfischer: I'm not sure what the preferred style is here, I just hacked this together by looking at other completions in this project (ex. `_nvm`). Let me know if there's a better example to follow.

src/_emulator
syohex: I suppose this function can be simplified as below

```bash
_list_avds() {
  local -a _avds=($HOME/.android/avd/*.ini(N.:t:r))
  echo "${_avds[@]}"
}
```

`N` glob qualifier: set null glob option. don't raise error if glob is not matched
`.` glob qualifier: match only files
`:t` modifier: Remove directory same as `basename` command
`:r` modifier: Remove file extension.

I suppose it is better to use builtin features rather than using external command(in this case `ls` command)

nfischer: Thanks, that's a lot better!

syohex:
I don't understand android emulator command line tool well. If it should complete android virtual device name only typing `@` character, you can write it as below. However it is not necessary as far as I read its document(https://developer.android.com/studio/run/emulator-commandline)

```diff
diff --git a/src/_emulator b/src/_emulator
index 265dccd..d768e0b 100644
--- a/src/_emulator
+++ b/src/_emulator
@@ -119,10 +119,9 @@ _arguments \
 case $state in
   cmds)
     local -a _avds=($(_list_avds))
-    for ((i=1; i<=${#_avds[@]}; i++)); do
-      _avds[i]="@${_avds[i]}"
-    done
-    _values 'avds' "${_avds[@]}"
+    if compset -P '@'; then
+      _values 'avds' "${_avds[@]}"
+    fi
     ;;
 esac
```

nfischer:
Thanks!

The main observable difference is your suggestion won't start completing until the user types the "@" symbol, but this PR as merged will suggest AVDs without needing any starter text. I think the behavior-as-merged is just a bit more discoverable, and aligns with what I would expect as a user. But either could work.

PR #697, opened 2020-03-29
Add completion for git-revise

dancek:

### DIFF
--- a/src/_git-revise
+++ b/src/_git-revise
@@ -1,0 +1,52 @@
+#compdef git-revise
+#description update, split and rearrange commits
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for git-revise 0.5.1
+#  (https://github.com/mystor/git-revise/).
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * Hannu Hartikainen (https://github.com/dancek)
+#
+# ------------------------------------------------------------------------------
+
+__git-revise_commits() {
+  local -a commits
+  commits=(${(f)"$(git log -20 --pretty=format:'%h:%d %s' | sed 's/: /:/')"})
+  _describe -V 'commit' commits
+}
+
+__git-revise_branches() {
+  local -a branches
+  branches=(${(f)"$(git for-each-ref --format='%(refname:short)' refs/heads/)"})
+  _describe 'branch' branches
+}
+
+_git-revise() {
+  local curcontext="$curcontext" ret=1
+
+  _arguments -s \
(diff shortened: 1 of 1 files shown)


syohex:
Thanks for PR. However I suppose it is better to set exclude option as below. For example, `--autosquash` and `--no-autosquash`. If we specify `--autosquash` option, then we don't want to complete `--no-autosquash` option.

```diff
diff --git a/src/_git-revise b/src/_git-revise
index cc9f92c..850c1ca 100644
--- a/src/_git-revise
+++ b/src/_git-revise
@@ -31,18 +31,18 @@ _git-revise() {
   local curcontext="$curcontext" ret=1
 
   _arguments \
-    {-h,--help}'[show help message and exit]' \
+    '(- * 1)'{-h,--help}'[show help message and exit]' \
     '--ref[reference to update]: :__git-revise_branches' \
     '--reauthor[reset the author of the targeted commit]' \
-    '--version[show version number and exit]' \
-    {--edit,-e}'[edit commit message of targeted commit(s)]' \
-    '--autosquash[automatically apply fixup! and squash! commits to their targets]' \
-    '--no-autosquash[force disable revise.autoSquash behaviour]' \
+    '(- * 1)--version[show version number and exit]' \
+    '(--edit -e)'{--edit,-e}'[edit commit message of targeted commit(s)]' \
+    '(--no-autosquash)--autosquash[automatically apply fixup! and squash! commits to their targets]' \
+    '(--autosquash)--no-autosquash[force disable revise.autoSquash behaviour]' \
     '--no-index[ignore the index while rewriting history]' \
-    {--all,-a}'[stage all tracked files before running]' \
-    {--interactive,-i}'[interactively edit commit stack]' \
-    {--message,-m}'[specify commit message on command line]:MESSAGE:()' \
-    {--cut,-c}'[interactively cut a commit into two smaller commits]' \
+    '(--all -a)'{--all,-a}'[stage all tracked files before running]' \
+    '(--interactive -i)'{--interactive,-i}'[interactively edit commit stack]' \
+    '(--message -m)'{--message,-m}'[specify commit message on command line]:MESSAGE:()' \
+    '(--cut -c)'{--cut,-c}'[interactively cut a commit into two smaller commits]' \
     '1: :__git-revise_commits' && ret=0
```

dancek:
Thanks for the help! This is a good idea, I just didn't know how to do it. I'll update the commit.

dancek:
I re-read the relevant part of `zshcompsys` and wrote mutually exclusive groups as they're in the [git-revise source code](https://github.com/mystor/git-revise/blob/master/gitrevise/tui.py#L20-L85).

syohex:
Thanks

PR #702, opened 2020-04-13
Add toolbox completer

llua:
- [ x] This compdef is not already available in zsh.
- [ x] This compdef is not already available in their original project.
- [ x] I am the original author, or I have authorization to submit this work.
- [ x] It has a license header or I accept that it will be licensed under the terms of the Zsh license.

originally submitted it upstream, where it sat for awhile. https://github.com/containers/toolbox/pull/322

### DIFF
--- a/src/_toolbox
+++ b/src/_toolbox
@@ -1,0 +1,120 @@
+#compdef toolbox
+
+_toolbox-create() {
+  _arguments -C -A '-*' \
+    '--help[display help]' \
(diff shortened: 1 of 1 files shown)


syohex:
Please write license header like ⬇️ 

https://github.com/zsh-users/zsh-completions/blob/538f2a0287bf8d3775e3745e212e27ad29f5003d/src/_afew#L2-L39

PR #703, opened 2020-04-16
Make plugin compatible with zsh plugin standard

horosgrisa:
Make plugin compatible with zsh plugin standard
https://zdharma.org/Zsh-100-Commits-Club/Zsh-Plugin-Standard.html


syohex:
I suppose this change breaks many people configurations who does not use `zsh-completions` as plugin. Can symbolic link or other way be available instead of renaming directory ?

horosgrisa:
@syohex Done

okapia:
I'm not convinced that moving every function is an especially good idea. It also makes it messier to follow git history. I'm not even sure that the standard proposal is trying to enshrine the name "functions". Certainly it is good to encourage the use of function autoloading rather than inlining. For context, keep in mind that the document you're following is essentially a proposal by the author of one of the many plugin managers. There's useful suggestions there but I don't think the exact naming the functions directory is important. The concept of plugin managers is seemingly not popular among people on the zsh-workers list so they've not endorsed any standard. There are other conventions such as lowercase for descriptions that are documented in Etc/completion-style-guide. Many of these are not always followed by functions here.

PR #710, opened 2020-05-02
replace broken link to linux-mag with its latest archive.org backup

jsamr:


syohex:
thanks

PR #711, opened 2020-05-04
Update stack completions as in the current version 2.3.0.1 

kanashimia:
Stack actually provides [script for completions](https://docs.haskellstack.org/en/stable/shell_autocompletion/#shell-auto-completion), but it doesn't distinguish between 'no-' and standalone options, doesn't provide descriptions.

This PR was based on the output of `stack --info`, descriptions were slightly modified to decrease length.

### DIFF
--- a/src/_stack
+++ b/src/_stack
@@ -45,8 +45,10 @@ _stack () {
         --version'[show version]' \
         --numeric-version'[show only version number]' \
         --hpack-numeric-version"[show only hpack's version number]" \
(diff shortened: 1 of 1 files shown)


syohex:
@kanashimia Could you squash commits when you complete work ?

kanashimia:
@syohex yes, no problem.

nicoulaj:
Looks good to me, thanks!

PR #733, opened 2020-05-15
support cmake --install related options

zchrissirhcz:
According to cmake's documentation, starting from cmake 3.15, it support "--install" option.

I try to add "--install" related options according to existing "--build" options.


syohex:
BTW please write pull request comment based on pull request template and do not remove it

PR #748, opened 2020-06-03
pgsql: Check for pg_config existence

arp242:
pg_config isn't installed by default on many systems, and is usually
part of postgresql-libs-devel or some such, leading to needless
"pg_config not found" errors on completion.

### DIFF
--- a/src/_pgsql_utils
+++ b/src/_pgsql_utils
@@ -147,10 +147,9 @@ _pgsql_databases () {
     _pgsql_get_identity
 
     local _pgsql_services _pgsql_service_files
-    _pgsql_service_files=(
-      ~/.pg_service.conf
-      $(pg_config --sysconfdir)/pg_service.conf
-    )
+    _pgsql_service_files=(~/.pg_service.conf)
+    (( $+commands[pg_config] )) && _pgsql_service_files+=$(pg_config --sysconfdir)/pg_service.conf
+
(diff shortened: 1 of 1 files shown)


src/_pgsql_utils
```diff
@@ -147,10 +147,9 @@ _pgsql_databases () {
     _pgsql_get_identity
 
     local _pgsql_services _pgsql_service_files
-    _pgsql_service_files=(
-      ~/.pg_service.conf
-      $(pg_config --sysconfdir)/pg_service.conf
-    )
+    _pgsql_service_files=(~/.pg_service.conf)
+    [[ -x pg_config ]] && _pgsql_service_files+=$(pg_config --sysconfdir)/pg_service.conf
```
syohex, line 151: I think this checks that `pg_config` is in current directory and it is an executable file. I suppose this should check whether `pg_config` is installed as bellow.

```zsh
(( $+commands[pg_config] )) && _pgsql_service_files+=$(pg_config --sysconfdir)/pg_service.conf
```

Note: `(( $+commands[command_name]))` equals to `which command_name 1>/dev/null 2>/dev/null`

arp242: Oops, you're right; I misinterpreted the docs on this (and for some reason my brain still remembers the old tcsh behaviour where -X does look up in PATH). I already used `$+commands[..]` in my zshrc, but I had forgotten about it 😅

I've updated the commit, and tested both with and without pg_config existence to make sure it's correct.

syohex:
Thanks

PR #749, opened 2020-06-03
pgsql: add -X to psql commands to prevent loading ~/.psqlrc

arp242:
The ~/.psqlrc may alter the output. In particular, I have '\timing on'
in there to add timings. This is useful when using psql, but not when
relying on it in scripts.

For example:

    psql -Aqt -c "select n.nspname \
        from pg_catalog.pg_namespace n \
        where n.nspname "'!~'" '^pg_' \
          and n.nspname <> 'information_schema' \
        order by 1;" postgres 2>/dev/null

Gives the output:

    public
    Time: 1.882 ms

Whereas with -X it gives the standard output:

    public


syohex:
Thanks

Issue #682, opened 2020-03-04
Completion for `flutter` is outdated


drianoaz:
The support for Flutter cli tool completions was added to zsh-completions in mid-2018 (#565). Since then, several commands have been added in the flutter and these have not been updated.

Example of the new available commands:

  - **assemble**:                 Assemble and build flutter resources.
  - **attach**:                   Attach to a running application.
  - **bash-completion**:          Output command line shell completion setup scripts.
  - **emulators**:               List, launch and create emulators.
  - **pub**:                     Commands for managing Flutter packages.
  - **version**:                 List or switch flutter versions.

MaximDevoir:
Thank you @drianoaz for bringing this to our attention. I'm not familiar with the Flutter CLI. Do you think you would like to work on that?

drianoaz:
Yes, I would like to work on that. In the next few days, I will open a PR.

syohex:
I have merged #754. Please check latest version.

PR #754, opened 2020-07-20
Update new completions for Flutter

kamleshkc2002:
closes https://github.com/zsh-users/zsh-completions/issues/682

Add new completions for the Flutter CLI

Updated as per the original issue.

- [X] This compdef is not already available in zsh.
- [X] This is a finished work.
- [X] It has a license header or I accept that it will be licensed under the terms of the Zsh license.

### DIFF
--- a/src/_flutter
+++ b/src/_flutter
@@ -86,6 +86,21 @@ _arguments -C -A "-*" \
             '--no-preamble[When analyzing the flutter repository, display the number of files that will be analyzed. (defaults to on)]' \
           && ret=0
         ;;
+        (assemble)
+            _arguments -C \
+            '(-h --help)'{-h,--help}'[Print this usage information.]' \
+          && ret=0
+        ;;
+        (attach)
+            _arguments -C \
+            '(-h --help)'{-h,--help}'[Print this usage information.]' \
+          && ret=0
+        ;;
+        (bash-completion)
+            _arguments -C \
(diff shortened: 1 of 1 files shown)


src/_flutter
```diff
@@ -380,7 +404,12 @@ _root_commands() {
     'test:Run Flutter unit tests for the current project.'
     'trace:Start and stop tracing for a running Flutter app.'
     'update-packages:Update the packages inside the Flutter repo.'
-    'upgrade:Upgrade your copy of Flutter.')
+    'upgrade:Upgrade your copy of Flutter.'
+    'assemble:Assemble and build flutter resources.'
```
syohex, line 408: Original code sorts sub-command by alphabetical order. Could you insert new commands by alphabetical order and keep alphabetical order ?

src/_flutter
```diff
@@ -340,7 +340,31 @@ _arguments -C -A "-*" \
             '(-h --help)'{-h,--help}'[Print this usage information.]' \
           && ret=0
         ;;
-        
+        (assemble)
```
syohex, line 343: Same as other comment.

kamleshkc2002: sounds good.. updated.

src/_flutter
kamleshkc2002: sounds good. updated.

syohex:
Thanks for updating code.

Issue #762, opened 2020-09-14
[INFO] Flameshot will provide its own completions


TheAifam5:
Hey, while using the git version of flameshot on AUR I noticed that there is conflict for `_flameshot`.

The change might be in future releases:
https://github.com/flameshot-org/flameshot/commit/23ae6c15796a61a86a4df612dbb1990c19585eda

Best regards,
TheAifam5

Feresey:
anybody here?

syohex:
#764

syohex:
I have merged #764

PR #764, opened 2020-09-23
Remove flameshot completion

syohex:
It is provided from original project now

See #762. flameshot provides its own completion file now.


kapsh:
Sorry. It is me who added zsh support into flameshot's existing bash completion file because I never noticed that this project provides native one. If you think this one feels nicer to use, then it may be better to upstream it while removing here.

syohex:
> If you think this one feels nicer to use, then it may be better to upstream it while removing here.

I suppose both provides same features. So we should remove our version by our policy.

jetersen:
Could this be included in a new tag?

syohex:
> Could this be included in a new tag?

No

jetersen:
ah my issue was I had a leftover symlink inside my `~/.zinit/completions` 😅

PR #770, opened 2020-11-27
Add completion file for bat

xyproto:

It is available in the bat package for Arch Linux (which I maintain) but I think it would be a good idea to have it in the zsh-completions package instead, which uses this repo as the source.

Licensed under MIT + APACHE (text included as a comment at the top of the file).


syohex:
I'm afraid that we don't add a completion which is already provided by original project. Please use original version.

xyproto:
I see. Thanks for reviewing the pull request!

PR #774, opened 2020-12-31
_mssh: Handle error cases more gracefully

simnalamburt:
I've created mssh autocompletion yesterday, and today I'm making it deals with various corner cases more gradecully.

1.  Do nothing when there's no AWS CLI
2.  Print error message gracefully if AWS CLI failed

### DIFF
--- a/src/_mssh
+++ b/src/_mssh
@@ -39,6 +39,11 @@
 #
 # ------------------------------------------------------------------------------
 
+# Do nothing if there's no AWS CLI
+if (( ! $+commands[aws] )); then
+  return
+fi
+
 # Define function only when it doesn't exist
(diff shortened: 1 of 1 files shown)


nicoulaj:
I do not have setup for properly testing this, any other advice ?

simnalamburt:
Well.. even if you don't have working AWS CLI, no ugly unreadable error messages should appear after `mssh<tab>`. That's the test for the first diff.

And create some dummy aws binary which always fails like below, and put it under $PATH.

```
#!/bin/bash
>&2 echo "error"
exit 1
```

Then try `mssh<tab>` again. The error message should be organized properly like below:

image

PR #775, opened 2021-01-03
Fix completion of cipher algorithms

brainplot:
Tab-completing `openssl` would give an error about the `list-cipher-algorithms` being an invalid command. The correct command seems to be `list -cipher-algorithms` (notice the space).

openssl

### DIFF
--- a/src/_openssl
+++ b/src/_openssl
@@ -611,7 +611,9 @@ _openssl_gendsa() {
 _openssl_genpkey() {
   # written for openssl 1.0.1k
   local ciphers cipher_opts
-  ciphers=( ${$(openssl list-cipher-algorithms | cut -d' ' -f1)} )
+  if ! ciphers=( ${$(openssl list-cipher-algorithms | cut -d' ' -f1)} ) 2>/dev/null ; then
+    ciphers=( ${$(openssl list -cipher-algorithms | cut -d' ' -f1)} )
+  fi
   cipher_opts=()
   for alg in ${ciphers}; do
     cipher_opts=(${cipher_opts} "(${${(l:32:: ::-:)ciphers[@]}//  / })-${alg}[use this cipher to encrypt the key]")


syohex:
It works newer OpenSSL since 1.1.0. However old version OpenSSL such as macOS bundled one does not support `list -cipher-algorithms`. I'm not sure there is no problem for dropping older versions. I suppose it is better to support older and newer versions.

brainplot:
Oh, I see. I have mine upgraded with Homebrew. Should we just close this then?

syohex:
How about a following patch ?

```diff
diff --git a/src/_openssl b/src/_openssl
index 4a44ea0..9612b6a 100644
--- a/src/_openssl
+++ b/src/_openssl
@@ -611,7 +611,11 @@ _openssl_gendsa() {
 _openssl_genpkey() {
   # written for openssl 1.0.1k
   local ciphers cipher_opts
-  ciphers=( ${$(openssl list-cipher-algorithms | cut -d' ' -f1)} )
+  if openssl list-cipher-algorithms >/dev/null 2>/dev/null ; then
+    ciphers=( ${$(openssl list-cipher-algorithms | cut -d' ' -f1)} )
+  else
+    ciphers=( ${$(openssl list -cipher-algorithms | cut -d' ' -f1)} )
+  fi
   cipher_opts=()
   for alg in ${ciphers}; do
     cipher_opts=(${cipher_opts} "(${${(l:32:: ::-:)ciphers[@]}//  / })-${alg}[use this cipher to encrypt the key]")
```

brainplot:
LGTM. I'll apply it and force-push!

brainplot:
I slightly changed the logic such that `list-cipher-algorithms` isn't run twice if the command does exist. Let me know if there are any problems with this change!

nicoulaj:
LGTM, thanks !

As @syohex said, the best is to support older versions, but very few completions actually do it, and we usually just update the compdef to the latest version.

PR #776, opened 2021-01-16
Use correct makefile for target completion

askebm:
cmake targets completion looked for a Makefile not in the build directory.
 It does now.


syohex:
thanks

PR #777, opened 2021-01-18
udisksctl: fix paths with a colon

dancek:
For paths such as `/dev/disk/by-id/usb-JetFlash_Transcend_64GB_15AFGA849642-0:0`, `_describe` parses the colon as field separator. This results in wrong completions and the general feeling that udisksctl is broken because it won't mount my pendrive.

Since the output of `udisksctl complete` is (apparently) just a list of paths, it's safe to escape all colons with a backslash.

### DIFF
--- a/src/_udisksctl
+++ b/src/_udisksctl
@@ -52,7 +52,7 @@
 _paths() {
     local -a _path_list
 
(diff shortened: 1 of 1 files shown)


syohex:
Thanks

PR #784, opened 2021-02-27
add 'shellcheck' completion

Z5483:

### DIFF
--- a/src/_shellcheck
+++ b/src/_shellcheck
@@ -1,0 +1,65 @@
+#compdef shellcheck
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 Github zsh-users - http://github.com/zsh-users
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the zsh-users nor the
+#       names of its contributors may be used to endorse or promote products
+#       derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL ZSH-USERS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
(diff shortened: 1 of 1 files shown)


syohex:
How about like this without variable and state ? Global variable definitions affects globally and it is not preferable.

```zsh
_arguments -C \
  {-a,--check-sourced}'[Include warnings from sourced file]' \
  {-C,--color=}'[Use color]: :(auto always never)' \
  {-i,--include=}'[Consider only given types of warnings]:specify code' \
  {-e,--exclude=}'[Exclude types of warnings]:specify code' \
  {-f,--format}'[Output format]: :(checkstyle diff gcc json json1 quiet tty)' \
  '--list-optional[List checks disabled by default]' \
  '--norc[Do not look for .shellcheckrc files]' \
  {-o,--enable}'[List of optional checks to enable (or '\''all'\'')]' \
  {-P,--source-path=}'[Specify path when looking for sourced files]:_files -/' \
  {-s,--shell=}'[Specify dialect]: :(sh bash dash ksh)' \
  {-S,--severity=}'[Minimum severity of errors to consider]: :(error warning info style)' \
  {-V,--version}'[Print version information]' \
  {-W,--wiki-link-count=}'[The number of wiki links to show, when applicable]:specify number' \
  {-x,--external-sources}'[Allow outside sources]' \
  '--help[Show this usage summary and exit]' \
  '*: :_files'
```

And please put a license header.

okapia:
> And please put a license header.

Given that "It has a license header or I accept that it will be licensed under the terms of the Zsh license." has been ticked, isn't that confirmation that Z5483 is happy for it to be distributed under the terms of the Zsh licence. Maybe the contribution header needs to force that to be more explicit.

The Zsh license should really always be the preferred choice because it is the only choice that makes it possible for zsh itself to include the completion function. There are fewer of them here precisely because they then often did then move to zsh itself. No header means the default choice - the Zsh licence.

okapia:
> `_arguments -C \`

Without the states, `-C` won't be needed any more.
Also, check whether `-s` or `-S` or `-A` options are applicable. `-s` indicates that short options can be combined in a single word (e.g `-aWx`), `-S` indicates that `--` ends options to allow a file with a name starting with `-` and `-A` is used with a pattern to indicate that the first matching argument ends options (BSD rather than GNU style option parsers).

>`   {-a,--check-sourced}'[Include warnings from sourced file]' \`

Zsh convention is for lowercase in descriptions: "include…"
That applies throughout.

>   `{-C,--color=}'[Use color]: :(auto always never)' \`

This has lost the description from the original patch: the word color should appear in the blank between the colons. Eliding that description may be common where we call a function or state that adds one but when listing matches, as here, it is best included.

>`   '--norc[Do not look for .shellcheckrc files]' \`
>`   {-o,--enable}'[List of optional checks to enable (or '\''all'\'')]' \`

I find the quoting rather less messy by using double quotes when I want single quotes in a description. Single quotes inside of double quotes need no extra quoting. On the first of those lines, I'd even use the shorter "don't" for the wording. I find it less ugly this way:

```
   "--norc[don't look for .shellcheckrc files]" \
   {-o,--enable}"[list of optional checks to enable (or 'all')]" \
```

>`   {-S,--severity=}'[Minimum severity of errors to consider]: :(error warning info style)' \`
>`   {-W,--wiki-link-count=}'[The number of wiki links to show, when applicable]:specify number' \`

Use the imperitive mood form for per-match descriptions, i.e. add "specify" or "give" to the beginning of these descriptions. The existing "specify number" description should just be "number". That's used as a heading.

Aside from that, the function lacks exclusion lists on the options. These mark the options that are mutually exclusive. That isn't necessary as such but presenting fewer matches makes completion more useful.

Z5483:
thank you for all the feedback!
I will fix the errors when I have time. Sorry for my sloppy PR!

syohex (approved):
Looks good

PR #792, opened 2021-04-19
added completions for opus-tools (opusenc, opusdec & opusinfo)

kugland:

### DIFF
--- a/src/_opustools
+++ b/src/_opustools
@@ -1,0 +1,112 @@
+#compdef opusenc opusdec opusinfo
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 Andre Kugland, MIT License
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for opus-tools (https://opus-codec.org/)
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * Andre Kugland (https://github.com/kugland)
+#
+# ------------------------------------------------------------------------------
+
+case $service in
+  opusenc)
+    _arguments \
+      '1:input file:_files -g "*.(#i)(wav|flac|ogg|aif(|f))(-.)"' \
+      '2:output file:_files -g "*.(#i)opus(-.)"' \
+      '(- 1 *)'{-h,--help}'[Show usage information]' \
+      '(- 1 *)'{-V,--version}'[Show version information]' \
+      '(- 1 *)--help-picture[Show help on attaching album art]' \
+      '--quiet[Enable quiet mode]' \
+      '--bitrate[Set target bitrate in kbit/s (6-256/channel)]:target bitrate in kbit/s (6-256/channel)' \
+      '--comp[Set encoding complexity (0-10)]:encoding complexity (1-10) [10]:(1 2 3 4 5 6 7 8 9 10)' \
+      '--cvbr[Use constrained variable bitrate encoding]' \
+      '--downmix-mono[Downmix to mono]' \
+      '--downmix-stereo[Downmix to stereo (if >2 channels)]' \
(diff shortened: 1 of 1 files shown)


syohex:
It looks good. Thanks

src/_opustools
```diff
@@ -0,0 +1,112 @@
...
+      '--max-delay[Set maximum container delay in milliseconds (0-1000) \[1000\]]:maximum container delay in milliseconds (0-1000) [1000]' \
+      '--music[Tune low bitrates for music (override automatic detection)]' \
+      '--no-phase-inv[Disable use of phase inversion for intensity stereo]' \
+      '--speech[Tune low bitrates for speech (override automatic detection)]' \
+      '--vbr[Use variable bitrate encoding (default)]' \
+      '--album[Set album or collection]:album or collection' \
+      '--date[Set date of track (YYYY, YYYY-MM, or YYYY-MM-DD)]:date of track (YYYY, YYYY-MM, or YYYY-MM-DD)' \
+      '--discard-comments[Don'\''t keep metadata when transcoding]' \
+      '--discard-pictures[Don'\''t keep pictures when transcoding]' \
+      '--padding[Reserve n extra bytes for metadata \[512\]]:padding size in bytes' \
```
okapia, line 62: Convention is to put the default in the per-match description and to uses parentheses for units:
```
'--padding[reserve extra bytes for metadata]:padding size (bytes) [512]' \
```
I'd also remove the _n_ placeholder because whatever usage statement it is cross-referenence to did not also come over when copying and pasting from documentation.

src/_opustools
```diff
@@ -0,0 +1,112 @@
+#compdef opusenc opusdec opusinfo
+# ------------------------------------------------------------------------------
+# Copyright (c) 2021 Andre Kugland, MIT License
```
okapia, line 3: Given the ZSH rather than the MIT licence, I would very happily apply this to the zsh distribution itself which would allow it to reach more users. That's up to you of course.

src/_opustools
```diff
@@ -0,0 +1,112 @@
...
+#  * Andre Kugland (https://github.com/kugland)
+#
+# ------------------------------------------------------------------------------
+
+case $service in
+  opusenc)
+    _arguments \
+      '1:input file:_files -g "*.(#i)(wav|flac|ogg|aif(|f))(-.)"' \
+      '2:output file:_files -g "*.(#i)opus(-.)"' \
+      '(- 1 *)'{-h,--help}'[Show usage information]' \
```
okapia, line 41: Zsh convention is not to capitalise descriptions.

src/_opustools
```diff
@@ -0,0 +1,112 @@
...
+      '--expect-loss[Set expected packet loss in percent]:expected packet loss in percent [0]' \
+      '--framesize[Set maximum frame size in milliseconds \[20\]]:maximum frame size in milliseconds [20]:(2.5 5 10 20 40 60)' \
+      '--hard-cbr[Use hard constant bitrate encoding]' \
+      '--max-delay[Set maximum container delay in milliseconds (0-1000) \[1000\]]:maximum container delay in milliseconds (0-1000) [1000]' \
+      '--music[Tune low bitrates for music (override automatic detection)]' \
+      '--no-phase-inv[Disable use of phase inversion for intensity stereo]' \
+      '--speech[Tune low bitrates for speech (override automatic detection)]' \
+      '--vbr[Use variable bitrate encoding (default)]' \
+      '--album[Set album or collection]:album or collection' \
+      '--date[Set date of track (YYYY, YYYY-MM, or YYYY-MM-DD)]:date of track (YYYY, YYYY-MM, or YYYY-MM-DD)' \
```
okapia, line 59: You could call `_dates -f "%F"`. If you don't use menu selection, `_dates` might seem fairly pointless but with it, you get a calendar.

src/_opustools
```diff
@@ -0,0 +1,112 @@
...
+      '--raw-chan[Set number of channels for raw input \[2\]]:number of channels' \
+      '--raw-endianness[Set endianness for raw input, 0 for little endian, 1 for big \[0\]]:endianness:(("0"\:"big endian" "1"\:"little endian"))' \
+      '--raw-rate[Set sampling rate for raw input \[48000\]]:sample rate in Hz' \
+      '--raw[Interpret input as raw PCM data without headers]' \
+      '--serial[Force use of a specific stream serial number]:stream serial number' \
+      '--save-range[Save check values for every frame to a file]:file:_files' \
+      '--set-ctl-int[Pass encoder control]:encoder control'
+    ;;
+  opusdec)
+    _arguments \
```
okapia, line 80: I take it you verified that `-S` was not applicable here? Does `--` terminate options? What about other options?

src/_opustools
```diff
@@ -0,0 +1,112 @@
...
+      '--force-wav[Force Wave header on output]' \
+      '--packet-loss[Simulate n% random packet loss]:percentage of packets' \
+      '--save-range[Save check values for every frame to a file]:output:_files'
+    ;;
+  opusinfo)
+    _arguments \
+      '(- 1 *)-h[Show usage information]' \
+      '(- 1 *)-V[Show version information]' \
+      '-q[Make less verbose]' \
+      '-v[Make more verbose]' \
```
okapia, line 101: make is a transitive verb so those descriptions are fairly horrible. Also, are they mutually exclusive because what they claim to do looks opposite. If so, add exclusion lists, e.g. `(-v)`. If one is only asserting the default, consider prefixing with `!` to make it ignored by default.

okapia (commented):
The patch has already been pushed but I do have some comments so have reviewed it anyway.

src/_opustools
kugland: I can use the MIT licence, sure. I was looking into the code of the completions included in the zsh distribution, and they have no licence notice. Should I just remove the notice from this file too then?

src/_opustools
kugland: What is the convention for the range of acceptable values?

okapia: Normally that is presented much like the units, so e.g:

```
'--max-delay[set maximum container delay]:delay (milliseconds) (0-1000) [1000]' \
```

src/_opustools
okapia: In both zsh and this project, the LICENSE file at the top-level is the zsh licence and it applies to any file that doesn't have some other exception notice. For this project, there are a few examples like _nano and _virtualbox without a different licence but there are relatively few because they were often moved to zsh. In zsh, there are only a couple with other licences. It'd be best to correct for this project and for a later import to zsh, headers would be stripped and attribution recorded in the ChangeLog/git history (but I would handle that).

PR #793, opened 2021-04-19
fixed issues raised in the review of opus-tools completion script

kugland:
* Changed all descriptions to start with lower case.
* Using conventional formatting for arg ranges and types.
* Removed copyright notice: now it should be licensable with Zsh license.
* -v and -q in opusinfo can be used multiple times and are not mutually exclusive.
* Added -S to _arguments for opusenc and opusdec, and -S -s for opusinfo.

### DIFF
--- a/src/_opustools
+++ b/src/_opustools
@@ -1,104 +1,71 @@
 #compdef opusenc opusdec opusinfo
-# ------------------------------------------------------------------------------
-# Copyright (c) 2021 Andre Kugland, MIT License
-#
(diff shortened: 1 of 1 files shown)


okapia:
I've merged this into zsh itself in commit 92a7b7f4.

I think this project prefers to keep the basic header with author and description.
At least once zsh 5.9 is out we'll want to remove it from here (along with some others) but it does no harm having it here until then.

syohex:
I apologize and thanks for your fix.

PR #794, opened 2021-04-21
Add license header to #793

syohex:
This is related to #793

### DIFF
--- a/src/_opustools
+++ b/src/_opustools
@@ -1,104 +1,71 @@
 #compdef opusenc opusdec opusinfo
-# ------------------------------------------------------------------------------
-# Copyright (c) 2021 Andre Kugland, MIT License
-#
-# Permission is hereby granted, free of charge, to any person obtaining a copy
-# of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/_opustools
```diff
@@ -1,5 +1,44 @@
...
+#  Completion script for opus-tools (https://opus-codec.org/)
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * Andre Kugland (https://github.com/kugland)
+#
+# ------------------------------------------------------------------------------
+
```
okapia, line 41: @syohex You do realise that this is a 3-clause BSD licence and NOT the zsh licence.

My understanding is that the absence of a licence header indicates that the zsh licence is to be used. Note how this applies to many functions originally written by Julien Nicoulaud who originated this project. There aren't many because many got merged into zsh. See _virtualbox and _android for examples.

For the zsh licence, see the file named LICENSE in the top-level of either this project or zsh.

syohex: Sorry I misunderstood. I'll fix it later

PR #799, opened 2021-04-30
fix license

syohex:
https://github.com/zsh-users/zsh-completions/pull/794#discussion_r623431843

@okapia


okapia:
Looks good, thanks!

PR #800, opened 2021-05-02
gist -u or gist --update will auto-complete user gists

maniac-en:
This commit will help the user to easily update their existing gist
using zsh auto-completion rather than going to github and then
extracting the required gist hash/URL

Note: This commit effects the working of file: `src/_gist`


syohex:
thanks

PR #808, opened 2021-07-15
Add completion for rmlint

oxiedi:

### DIFF
--- a/src/_rmlint.sh
+++ b/src/_rmlint.sh
@@ -1,0 +1,42 @@
+#compdef rmlint.sh -P rmlint.*.sh
+
+# Copyright (c) 2021 Github zsh-users - http://github.com/zsh-users
+#
+# Permission is hereby granted, without written agreement and without
+# licence or royalty fees, to use, copy, modify, and distribute this
+# software and to distribute modified versions of this software for any
+# purpose, provided that the above copyright notice and the following
+# two paragraphs appear in all copies of this software.
+#
+# In no event shall the Zsh Development Group be liable to any party for
+# direct, indirect, special, incidental, or consequential damages arising out
+# of the use of this software and its documentation, even if the Zsh
+# Development Group have been advised of the possibility of such damage.
+#
+# The Zsh Development Group specifically disclaim any warranties, including,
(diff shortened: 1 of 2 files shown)


src/_rmlint.sh
```diff
@@ -0,0 +1,13 @@
+#compdef rmlint.sh -P rmlint.*.sh
```
syohex, line 1: Is this file necessary ? I installed `rmlint` but there is no `rmlint.sh` file.

oxiedi: This is produced by `rmlint` itself:

    % rmlint >/dev/null
    % ./rmlint.sh -h
    usage: ./rmlint.sh OPTIONS

    OPTIONS:

      -h   Show this message.
      -d   Do not ask before running.
      -x   Keep rmlint.sh; do not autodelete it.
      -p   Recheck that files are still identical before removing duplicates.
      -r   Allow deduplication of files on read-only btrfs snapshots. (requires sudo)
      -n   Do not perform any modifications, just print what would be done. (implies -d and -x)
      -c   Clean up empty directories while deleting duplicates.
      -q   Do not show progress.
      -k   Keep the timestamp of directories when removing duplicates.
      -i   Ask before deleting each file
    % rmlint --backup >/dev/null
    % ./rmlint.2021-07-16T12:00:00Z.sh -h
    ...

syohex: Thanks for information. I'll check it later

syohex:
@oxiedi Your completion files looks good to me. Could you add license header to completion files ?

oxiedi:
> Could you add license header to completion files ?

Doesn't lack of the license header mean I'm OK with the Zsh license (I am, BTW)? Should I include the Zsh license in the header then?

syohex:
> Should I include the Zsh license in the header then?

Yes, please.

oxiedi:
Done.

> It has a license header or I accept that it will be licensed under the terms of the Zsh license.

This point is misleading then.

syohex:
Thanks for your work.

> This point is misleading then.

Sorry. I'll try to update pull request template later.

PR #815, opened 2021-08-23
xinput: escape colons in device names

yut23:
The completion menu breaks if any device names have colons in them, as colons are used to separate completion matches from their descriptions.


syohex:
thanks

PR #825, opened 2021-12-01
fix: entrypoint fpath expansion

MunifTanjim:
If I'm in `zsh-completions` directory and I source `zsh-completions.plugin.zsh`,

`"${0:h}/src"` was expnaded to `./src`, which is not what we want.

The updated `"${0:A:h}/src"` is expanded to `/full/absolute/path/to/zsh-completions/src`.


syohex:
Thanks

Issue #802, opened 2021-05-22
Yarn v2 workspace autocomplete not working


phuctm97:
I'm using yarn v2 and looks like workspace auto-completion is not working correctly. Any idea on how can it be solved, would love to contribute if possible!

Thank you.

phuctm97:
Just went a little into it, it is caused by: `yarn workspaces info` is not valid in Yarn v2, instead it's `yarn workspace list --json`, the output is a bit different, too.

kachkaev:
@phuctm97 have you found any solution for Yarn 2/3? Just wonder if there is any other source of autocompletions.

Seems like the problem is not only related to workspaces. After I have migrated https://github.com/kachkaev/njt and https://github.com/kachkaev/suppress-exit-code to Yarn Berry, I am no longer seeing suggestions for `package.json` scripts in both projects. I.e. when I type `yarn fi [tab]`, it does not turn into `yarn fix`. Path autocompletion is also lost: `yarn test sr [tab]` does not become `yarn test src/`.

Interestingly, autocompletion for default commands still works, i.e. `yarn ini [tab]` becomes `yarn init`.

syohex:
Sorry very late reply. I'm not sure, is https://github.com/zsh-users/zsh-completions/pull/827 ok ?

![yarn](https://user-images.githubusercontent.com/554281/145044796-1435f5be-f6cd-47f0-9439-ae50b551ca30.gif)

syohex:
I have merged #827. Please check latest version and close this issue if you have no problems,

kachkaev:
Thanks @syohex and sorry for a slow reply. Workspace autocompletion seems to work for me both with Yarn 1 and 3 👍 Just got the most recent version on a new laptop and could finally test it. Great stuff!

PR #827, opened 2021-12-07
Update collect workspaces command

syohex:

This is related to #802

### DIFF
--- a/src/_yarn
+++ b/src/_yarn
@@ -100,22 +100,24 @@ _yarn_commands_scripts() {
 }
 
 _yarn_scripts() {
-  local -a binaries scriptNames scriptCommands scripts
+  local -a binaries scripts
   local -a commands
 
   if [[ -n $_yarn_run_cwd ]]; then
-    scriptNames=($(cd $opt_args[--cwd] && cat package.json | perl -0777 -MJSON::PP -n -E '%r=%{decode_json($_)->{scripts}}; printf qq($_:$r{$_}\n) for sort keys %r'))
-    scriptCommands=("${(@f)$(cd $opt_args[--cwd] && cat package.json | perl -0777 -MJSON::PP -n -E '%r=%{decode_json($_)->{scripts}}; printf "$r{$_}\n" for sort keys %r')}")
-    binaries=($(cd $opt_args[--cwd] && echo node_modules/.bin/*(x:t)))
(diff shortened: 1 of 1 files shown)


kachkaev:
Thanks for looking into this @syohex!

The change might be breaking of Yarn v1 users. We should probably use `yarn --version` in some kind of an `if` statement to avoid that. There are also calls to `yarn run --json`, which don’t work in Berry.

Broadly speaking, I wonder if the difference between Yarn Classic and Yarn berry is big enough to implement autocompletion in a separate file. So conceptually, there can be a single `if` statement triages between one mode or the other.

syohex:
How about https://github.com/zsh-users/zsh-completions/pull/827/commits/2f54f2b21287a5753728feba1ae2abcbf9df2aab ?

kachkaev:
I’m not an expert in shell scripts, but this looks like a step in the right direction 👍 Something similar should probably be done for calls to `yarn run --json` too. There may be more discrepancies, but `yarn workspaces info` & `yarn run --json` are only ones I’ve spotted so far.

syohex:
I'm not familiar with yarn. Are there any alternative commands same as `yarn run --json` in yarn v2 or higher versions ? If there is not such command, it is necessary to parse package.json like `cat package.json | perl -0777 -MJSON::PP -n -E '$r = decode_json($_); say for keys %{$r->{scripts}}'`

kachkaev:
I am still new to Yarn berry myself, so am not sure if my answer is 100% right. The `yarn run` command was added in https://github.com/yarnpkg/berry/pull/290 but I could not find `--json` there.

It might be possible to run `yarn run` and `run bin` and parse the output. It looks very similar for both commands.

syohex:
I have updated without `yarn run --json` at https://github.com/zsh-users/zsh-completions/pull/827/commits/839df669222b7c25b8ffcacbcbd3f01c9154df24

kachkaev:
Cool, thank you @syohex! I can try checking the new script over the weekend if time allows!

Looking into `node_modules/.bin` should work for most folks on both yarn versions. The only problem I can imagine is to do with the usage of [Yarn PnP](https://yarnpkg.com/features/pnp), which bypasses the creation of `node_modules` directory. It is enabled in Yarn v2 and v3 by default.

From my experience, Yarn PnP is still experimental and so has compatibility issues with some popular tools. Going forward, it’d be best to leverage `yarn run` and `yarn bin` because it’s the most robust source of truth possible, but I don’t see reasons why we can’t enable Yarn Berry + `node_modules` only as a mid step 🙂

syohex:
I have update for no `node_modules` case https://github.com/zsh-users/zsh-completions/pull/827/commits/790890e25658f9b5c6ba68264d888a2b191b895e

PR #832, opened 2021-12-26
Add completions for spt (spotify-tui)

2KAbhishek:
https://github.com/Rigellute/spotify-tui

### DIFF
--- a/src/_spt
+++ b/src/_spt
@@ -1,0 +1,352 @@
+#compdef spt
+
+autoload -U is-at-least
+
+_spt() {
+    typeset -A opt_args
+    typeset -a _arguments_options
+    local ret=1
+
+    if is-at-least 5.2; then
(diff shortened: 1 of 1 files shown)


syohex:
`spt` already provides zsh completion.  We can generate it by `spt --completion zsh`.

2KAbhishek:
@syohex, we have to manually add it to completions once `sp` generates the script, having the completions here will be a bit more convenient don't you think?

okapia:
@2KAbhishek Given that the file you submitted is a copy of the output from `spt --completion zsh`, you can't claim to be the author of this: you don't own the copyright.

Where upstream projects include a zsh completion, the people who package it for distributions should take care to install the completion in `$PREFIX/share/zsh/site-functions`. If they don't on your system, raise a bug or pull-request with them.

The actual function also leaves a lot to be desired in terms of quality.

Issue #603, opened 2018-10-23
oh-my-zsh installation instructions are sub-optimal


segevfiner:
The current instructions: [README.md#oh-my-zsh](https://github.com/zsh-users/zsh-completions/blob/5bef0323984bbec2def9839fb52db00285a07ebf/README.md#oh-my-zsh), suggest to add zsh-completions to `plugins` and rerun `compinit`.

But that's sub-optimal. This will cause Zsh to rebuild the completion cache **twice** on each Zsh invocation: The first when you source oh-my-zsh, and the second time when you rerun `compinit`. This defeats the whole point of the completions cache, and will negatively impact the startup time of Zsh.

Sadly, oh-my-zsh doesn't seem to have a way for a plugin to add more `fpath` entries, besides the root folder of the plugin, before it invokes `compinit`. I think it's best to just suggest adding the zsh-completions `src` dir to `fpath` before the `source $ZSH/oh-my-zsh.sh` line, in order to avoid this issue. Something like this:

```sh
fpath=($ZSH/custom/plugins/zsh-completions/src $fpath)
```

felipecrs:
Just to clarify, the suggestion is to replace:

```zsh
plugins=(… zsh-completions)

source "$ZSH/oh-my-zsh.sh"

autoload -U compinit && compinit
```

With:

```zsh
fpath+="${ZSH_CUSTOM:-"$ZSH/custom"}/plugins/zsh-completions/src"
```

The above is just an optimized version of the previously suggested, which takes the ZSH_CUSTOM into account if set.

References https://github.com/ohmyzsh/ohmyzsh/issues/10412

felipecrs:
It's important to remove `zsh-completions` from the OMZ plugins because otherwise fpath would be filled with the same contents twice.

ragnarok22:
I used to have my configuration like this on Arch with Kitty terminal:

```
plugins=(git zsh-completions zsh-autosuggestions zsh-syntax-highlighting)

fpath+=${ZSH_CUSTOM:-${ZSH:-~/.oh-my-zsh}/custom}/plugins/zsh-completions/src
source $ZSH/oh-my-zsh.sh
```

and I recently migrated to Mac and partial autocompletions doesn't work (Alt + right Arrow). I've been messing around with the plugin and reading the docs and issues and I just remove it from the plugin list and it works perfectly.

This is my final version:

```
plugins=(git zsh-autosuggestions zsh-syntax-highlighting)

[...]

PR #835, opened 2022-01-05
Use OMZ completions-dir.txt to improve installation

felipecrs:
This implements https://github.com/ohmyzsh/ohmyzsh/pull/10565.

Fixes https://github.com/zsh-users/zsh-completions/issues/603.


felipecrs:
Waiting for #603 before marking as ready to review.

felipecrs:
My last commit addresses the concern raised in https://github.com/ohmyzsh/ohmyzsh/issues/10412#issuecomment-967214321, where multiple entries of `src` could end up being added to `fpath`.

The approach I used was taken from https://unix.stackexchange.com/a/62581/427940, and I would really like to know if there better or cleaner approaches out there for handling such a situation in ZSH.

nicoulaj:
Hi, I understand this `completions-dir.txt` file is specific to OMZ ? I don't think this project should be polluted with this kind of plugin manager specific files, it just makes it harder for other packagers and users. Not a big deal, just my two cents.

felipecrs:
Specific for now, this is a new feature that I'm introducing at https://github.com/ohmyzsh/ohmyzsh/pull/10565. Other package managers can also make use of it if they want.

Also, isn't https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions.plugin.zsh also specific to OMZ?

nicoulaj:
The `<foo>.plugin.zsh` is the only convention that is widely adopted by "plugin managers".

Does anyone else have thoughts on this ?

felipecrs:
@nicoulaj just to mention, I totally understand your reasonings.

I'm just trying to solve an outstanding incompatibility issue between this project and OMZ. But any better ideas are 100% welcome, of course.

mcornella:
I agree, this solution is not elegant and any changes that we make need to be adopted by the majority of plugin managers. It seems that most PMs support changing `$fpath` in the `.plugin.zsh` file.

This is not taken into account in OMZ because the `compinit` call happens before the plugin.zsh file is read (this is to allow running `compdef` in the .plugin.zsh, most outstandingly used in the git plugin).

If there is a solution to allow support for completion changes in the .plugin.zsh file while running `compinit` later, this solution wouldn't even be needed. Other PMs seem to do this, so it looks to me like there's no need for this and we simply have to improve OMZ.

PR #837, opened 2022-01-17
git-flow: Adding description to be picked up.

seivan:
Will describe what `flow` is, as a git sub command.
Current state
image

Changes:
image
 

- [x] I have authorization to submit this work.


syohex:
Thanks

PR #843, opened 2022-02-06
Installation > git clone only last commit

AlfredJKwack:
When installing either manually or with oh-my-zsh it's better to create a shallow clone with a history truncated to the latest commit only. The installation will be faster and less resource intensive. End users don't care for the entire repo's history and branches anyway.


AlfredJKwack:
After reading through [the article](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/) it seems indeed clear that for the size and nature of the repo the costs will outweigh the benefits. 
Closing with this comment.

PR #849, opened 2022-02-22
fix presets

huwper:

existing check didn't seem to work with any version of cmake I tried. this does

### DIFF
--- a/src/_cmake
+++ b/src/_cmake
@@ -141,13 +141,21 @@ _cmake_generator_options() {
 # --------------
 (( $+functions[_cmake_presets] )) ||
 _cmake_presets() {
-  local invoke; invoke=(${words[@]})
-  # TODO: remove all arguments -* except -S
-  invoke[$CURRENT]=--list-presets
+
+  local parse_presets; parse_presets='
+import json    
+import os    
+try:    
+    with open("CMakePresets.json") as preset_file:    
+        print(os.linesep.join([ x["name"] + ":" + x.get("description", "") for x in json.load(preset_file)["configurePresets"]]))
+except:    
+    pass    
+
+'
 
   # TODO: Problems with quotes need eval
   # would need a way to exec the array
-  local list_presets; list_presets=(${(f)"$(eval "${invoke[@]} 2> /dev/null" | grep -Po '(?<=\").*(?=")' )"})
+  local list_presets; list_presets=(${(f)"$(eval python <<< $parse_presets)"})
 
   _describe 'presets' list_presets
 }
--- a/src/_cmake
+++ b/src/_cmake
@@ -147,7 +147,7 @@ _cmake_presets() {
 
   # TODO: Problems with quotes need eval
   # would need a way to exec the array
-  local list_presets; list_presets=(${(f)"$(eval "${invoke[@]} 2> /dev/null" | sed -n 's,^[[:space:]]*"\([^"]*\)"[[:space:]]*-[[:space:]]*\(.*\),\1:\2,p' )"})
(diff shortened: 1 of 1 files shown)


syohex:
Could you tell us what is problem, how to reproduce your issue ?

src/_cmake
```diff
@@ -147,7 +147,7 @@ _cmake_presets() {
 
   # TODO: Problems with quotes need eval
   # would need a way to exec the array
-  local list_presets; list_presets=(${(f)"$(eval "${invoke[@]} 2> /dev/null" | sed -n 's,^[[:space:]]*"\([^"]*\)"[[:space:]]*-[[:space:]]*\(.*\),\1:\2,p' )"})
+  local list_presets; list_presets=(${(f)"$(eval "${invoke[@]} 2> /dev/null" | grep -Po '(?<=\").*(?=")' )"})
```
syohex, line 150: BSD grep implementation like macOS, freeBSD doesn't implement `-P` option. Could you rewrite it with portable way ?

huwper:
> Could you tell us what is problem, how to reproduce your issue ?

the output of `cmake --list-presets` looks like this:
```
➜ cmake --list-presets 
Available configure presets:

  "default"
  "ninja-multi" - Ninja Multi-Config

```

My guess is the original implementation assumed there would always be a description. and therefore failed to parse values with no description (I don't know how sed works).

My new commit uses python to parse CMakePresets.json directly. This is what I think cmake expects developers of tools to do, cmake --list-presets is intended for users.

this is the output when I hit tab:
```
➜  example-presets cmake --preset=
default       -- 
ninja-multi   -- Default build using Ninja Multi-Config generator
```
notice it leaves the description field blank if there is none

huwper:
realised the new implementation doesn't check the -S flag passed to cmake so will only work if called from the source directory

syohex:
Thanks for explanation.

BTW can I rewrite it in Perl as below ? Because `zsh` and `zsh-completions` do not depend on python and I don't want to add python dependency.

```diff
diff --git a/src/_cmake b/src/_cmake
index a0b0a0e..2e8520a 100644
--- a/src/_cmake
+++ b/src/_cmake
@@ -141,13 +141,14 @@ _cmake_generator_options() {
 # --------------
 (( $+functions[_cmake_presets] )) ||
 _cmake_presets() {
-  local invoke; invoke=(${words[@]})
-  # TODO: remove all arguments -* except -S
-  invoke[$CURRENT]=--list-presets
-
   # TODO: Problems with quotes need eval
   # would need a way to exec the array
-  local list_presets; list_presets=(${(f)"$(eval "${invoke[@]} 2> /dev/null" | sed -n 's,^[[:space:]]*"\([^"]*\)"[[:space:]]*-[[:space:]]*\(.*\),\1:\2,p' )"})
+  local -a list_presets;
+
+  if [[ -e CMakePresets.json ]]; then
+    # strip UTF-8 BOM and then parse JSON
+    list_presets=("${(@f)$(sed '1s/^\xEF\xBB\xBF//' < CMakePresets.json | perl -0777 -MJSON::PP -n -E 'do{$k=$_;($e=$k)=~s/:/\\:/g; printf "$_->{name}:$_->{description}\n"} for @{decode_json($_)->{configurePresets}}')}")
+  fi
 
   _describe 'presets' list_presets
 }
```

syohex:
Thanks but I'll rewrite it later.

syohex:
I have merged #850. Thanks

nolange:
Well, that breaks the completion for me, as I also use zsh + cmake on systems without perl.

Also I took care that this works if you arent within the cmake source directory, you seem to assume that CMakePresets.json is in the CWD?

I am not fluent enough with zsh, perhaps you dont even need `sed`

Issue #852, opened 2022-03-02
git-revise: error outside git repository


ambroisie:
When trying to complete `git revise` arguments and options outside of a `git` repository, the message `fatal: not a git repository (or any of the parent directories): .git` is displayed over the cursor (which is ugly).

Could this be handled similarly to the way normal `git` commands are completed, like for example `git rebase` which just shows the `not a git repository` message under the command line.

PR #853, opened 2022-03-03
Don't complete if here is not git repository

syohex:

#852 CC: @ambroisie


ambroisie:
LGTM.

I believe the Zsh `git` completions do the following to show that the user is trying to use the completion outside of a git repository:

```zsh
  if (( ${#*:#0} > 0 )); then
    _message 'not a git repository'
    return 1
  fi
```

PR #861, opened 2022-04-26
Drop unsupported git protocol from Readme

saeltz:
When trying to clone the repository, git complained with:
```
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
```

Using HTTPS in the URL works just fine.


syohex:
Thanks

Issue #877, opened 2022-07-01
Go 1.18 new flags


Deleplace:
Hello, I'm trying `zsh-completions` for the first time, to use it with `go` subcommands and flags.

I can see that some recent options are missing, e.g.
`go test -fu<TAB>` ->  no completion, I would expect `go test -fuzz`

Shall I manually list the missing flags and open a PR to add them to [src/_golang](https://github.com/zsh-users/zsh-completions/blob/dd45d21d06118a70ac5b42a7968706108f6494dc/src/_golang#L645) ?

Thanks!

syohex:
Thanks for the suggestion. I have merged #878

PR #878, opened 2022-07-01
Add go 1.18 test flags

syohex:
And fix some typo

This is related to #877. CC: @Deleplace


syohex:
I forgot to add it. I have add at 15d077a

PR #868, opened 2022-06-16
Add Fig as an installation method to the README

ibayramli2001:
- [ ] This compdef is not already available in zsh.
- [ ] This compdef is not already available in their original project.
- [ ] I am the original author, or I have authorization to submit this work.
- [ ] This is a finished work.
- [ ] It has a header containing authors, status and origin of the script.
- [ ] It has a license header or I accept that it will be licensed under the terms of the Zsh license.

## PR Description

The [Fig Plugin Store](https://fig.io/plugins) supports 1-click install for 400+ shell plugins. We have over 100k users, thousands of whom manage their shell configuration with Fig.

Zsh completions is one of the most popular plugins on our store (it's on our front page!), so we'd really love to have Fig listed as a download method on your readme.

Thanks so much and please let me know if you have any questions!

PR #889, opened 2022-09-05
Add Fig as installation method

brendanfalk:
Just wanted to create a fresh thread for adding Fig as an installation method. Previous thread was here: https://github.com/zsh-users/zsh-completions/pull/868/files

I have updated the wording, taking into account all the commentary from the other threads.

Apologies again for the issues above - please let me know if any more questions/concerns! 

Brendan


bartekpacia:
Fig is dead now. I suggest to close this issue.

PR #891, opened 2022-09-09
make httpie completion work for both `http` and `https`

max-ae:

`httpie` provides both the `http` and `https` command. Let's make the completion work for both.


syohex:
Sorry for the very belated reply. Thanks for your PR.

PR #893, opened 2022-09-12
Utilize grep -E instead of deprecated/obsolescent egrep

rtizzy:
This PR changes the vagrant and virtualbox completions so they utilize `grep -E` instead of calling `egrep`.


syohex:
Thanks

PR #895, opened 2022-09-30
Remove _trash* because trashcli support completion

Freed-Wu:
Trashcli has its own shell completion which support bash/zsh/tcsh.

https://github.com/andreafrancia/trash-cli/blob/master/trashcli/shell_completion.py


syohex:
BTW how do we use trash own shell completion ? It looks that there is no documentation about it. Is it easy to migrate from zsh-completions to its own completion ?

Freed-Wu:
> there is no documentation about it

I notice it. It should be a forgiveness of the maintainer.

```sh
cmds=(trash-empty trash-list trash-restore trash-put trash)
for cmd in $cmds; do
  $cmd --print-completion bash | sudo tee /usr/share/bash-completion/completions/$cmd
  $cmd --print-completion zsh | sudo tee /usr/share/zsh/site-functions/_$cmd
  $cmd --print-completion tcsh | sudo tee /etc/profile.d/$cmd.completion.csh
done
```

So, `/usr/share/zsh/site-functions/trash*` must be removed to avoid conflicts.

Freed-Wu:
Can we add one section to `README.md` about how to use shell completion? @andreafrancia

andreafrancia:
Yes, sure just send me a pull request with the needed modifications for the README.

andreafrancia:
@Freed-Wu, I've merged your pull request.

syohex:
Thanks

PR #896, opened 2022-10-04
Fix nftables completion

huyz-git:
New version of `nft` doesn't allow put options after commands
```
# nft list ruleset -a
Error: syntax error, options must be specified before commands
nft list ruleset -a 
   ^             ~~
```
So the completion should follow this, otherwise tables/chains/rules cannot be completed.


syohex:
Thanks

PR #898, opened 2022-10-12
Update README.md

dpthegrey:
Adds step to install zsh-completions using Homebrew Formulae.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,32 @@ Note: adding it as a regular Oh My ZSH! plugin will not work properly (see [#603
 
 Add `zinit light zsh-users/zsh-completions` to your `~/.zshrc`.
 
+### Using Homebrew
+
+### [Brew](https://github.com/Homebrew/brew)
+* Install command:
+        
+        brew install zsh-completions
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,32 @@ Note: adding it as a regular Oh My ZSH! plugin will not work properly (see [#603
...
 Add `zinit light zsh-users/zsh-completions` to your `~/.zshrc`.
 
+### Using Homebrew
+
+### [Brew](https://github.com/Homebrew/brew)
+* Install command:
+        
+        brew install zsh-completions
+
+* To activate these completions, add the following to your `~/.zshrc`:
```
memark, line 57: Do we want to duplicate this information here, as it is the output by the `brew install` command the user would run?

dpthegrey: Came up to [this issue](https://github.com/Homebrew/homebrew-core/issues/108716) on homebrew and [a comment](https://github.com/Homebrew/homebrew-core/issues/108716#issuecomment-1224884682) states "The zsh completions formula and the homebrew completions are 2 separate things". What can be done in this case?

leopedroso45 (approved):

PR #985, opened 2023-01-18
Update completion for mix

vurden:
The Phoenix project was updated to add auth generator, figured it would be nice to have here as well.


vurden:
Sorry about all the renames, trying to make it line up with other merges.

syohex:
Thanks

PR #986, opened 2023-01-20
Add support for l3build

Freed-Wu:
See <https://github.com/latex3/l3build/pull/267>


syohex:
I'm afraid that our policy is that `Completions are not accepted when already available in their original project.`

See also
- https://github.com/zsh-users/zsh-completions/blob/master/CONTRIBUTING.md

PR #993, opened 2023-02-06
tmuxp: fix using workspace dirs other than ~/.tmuxp

infokiller:
For reference see https://github.com/tmux-python/tmuxp/blob/3ba64262839da6a643a19e70bb990ca3e589ad16/src/tmuxp/workspace/finders.py#L99-L128


syohex:
Thanks

PR #995, opened 2023-02-16
Add texdoc

Freed-Wu:


syohex:
I'm afraid that this PR is unacceptable for the reason same as #986.

https://github.com/zsh-users/zsh-completions/pull/986#issuecomment-1404821599

PR #1003, opened 2023-03-28
updated _bitcoin-cli to match bitcoind v24.1rc1

notmike-5:

### DIFF
--- a/src/_bitcoin-cli
+++ b/src/_bitcoin-cli
@@ -35,7 +35,7 @@
 # -------
 #
 #  * Ian Ker-Seymer (https://github.com/ianks)
-#
+#  * notmike
 # ------------------------------------------------------------------------------
 
 _bitcoin-cli() {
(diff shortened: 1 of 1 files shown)


src/_bitcoin-cli
```diff
@@ -35,27 +35,32 @@
...
 # ------------------------------------------------------------------------------
 
 _bitcoin-cli() {
   local context state line curcontext="$curcontext"
 
   _arguments -C \
     -?'[This help message]' \
+    -addrinfo'[Get the number of addresses known to the node, per network and total, after filtering for quality and recency. Total number of addresses known to the node may be higher.]' \
+    -chain='[Use the chain <chain> (default: main). Allowed values: main, test, signet, regtest]' \
+    -color='[Color setting for CLI output (default: auto). Valid values: always, auto, never]' \
```
syohex, line 48:
```suggestion
    -chain='[Use the chain <chain> (default: main). Allowed values: main, test, signet, regtest]:chain:(main test signet regtest)' \
    -color='[Color setting for CLI output (default: auto). Valid values: always, auto, never]:when:(always auto never)' \
```

src/_bitcoin-cli
```diff
@@ -35,27 +35,32 @@
...
     -conf='[Specify configuration file. Relative paths will be prefixed by datadir location. (default: bitcoin.conf)]:PATH:_files' \
     -datadir='[Specify data directory]:PATH:_directories' \
     -getinfo='[Get general information from the remote server.]' \
     -testnet'[Use the test chain]' \
     -regtest'[Enter regression test mode, which uses a special chain in which blocks can be solved instantly. This is intended for regression testing tools and app development.]' \
     -named'[Pass named instead of positional arguments (default: false)]' \
+    -netinfo'[Get network peer connection information from the remote server.]' \
```
syohex, line 55:
```suggestion
    -netinfo=-'[Get network peer connection information from the remote server.]::level_or_help:(0 1 2 3 4 help)' \
```

src/_bitcoin-cli
syohex: If the option can take specific arguments, we should set them as candidates

Reference
- https://zsh.sourceforge.io/Doc/Release/Completion-System.html

syohex:
Thanks

PR #1022, opened 2023-07-09
add trash-d completion

mirsella:

https://github.com/rushsteve1/trash-d

thanks !


src/_trash
```diff
@@ -0,0 +1,49 @@
...
+
+case $state in
+	(file)
+		(( CURRENT > 0 )) && line[CURRENT]=()
+		line=( ${line//(#m)[\[\]()\\*?#<>~\^\|]/\\$MATCH} )
+		_files -F line && ret=0
+		;;
+esac
+
+return $ret
```
syohex, line 49:
```suggestion
return $ret
# Local Variables:
# mode: Shell-Script
# sh-indentation: 2
# indent-tabs-mode: nil
# sh-basic-offset: 2
# End:
# vim: ft=zsh sw=2 ts=2 et
```

Please use same file format as other files. Use two spaces, use spaces instead of tabs and so on.

src/_trash
```diff
@@ -0,0 +1,49 @@
...
+# ------------------------------------------------------------------------------
+
+declare -a args opts=( -A '-*' )
+args=(
+	'(-d --dir)'{-d,--dir}'[Remove empty directories]'
+	'(-r --recursive)'{-r,--recursive}'[Delete directories and their contents]'
+	'(-v --verbose)'{-v,--verbose}'[Print more information]'
+	'(-i --interactive)'{-i,--interactive}'[Ask before each deletion]'
+	'(-I --interactive-once)'{-I,--interactive-once}'[Ask once if deleting 3 or more]'
+	'(-f --force)'{-f,--force}'[Don'\''t prompt and ignore errors]'
```
syohex, line 23:
```suggestion
	'(-i --interactive -I --interactive-once -f --force)'{-i,--interactive}'[Ask before each deletion]'
	'(-i --interactive -I --interactive-once -f --force)'{-I,--interactive-once}'[Ask once if deleting 3 or more]'
	'(-i --interactive -I --interactive-once -f --force)'{-f,--force}'[Don'\''t prompt and ignore errors]'
```

I think `-i`, `-I`, `-f` cannot be used at the same time.

src/_trash
```diff
@@ -0,0 +1,49 @@
...
+	'--version[Output the version and exit]'
```
syohex, line 24:
```suggestion
	'(- *)--version[Output the version and exit]'
```

If an user specifies `--version`, then it should not complete other options

src/_trash
```diff
@@ -0,0 +1,49 @@
...
+	'--list[List out the files in the trash]'
+	'--orphans[List orphaned files in the trash]'
+	'--restore[Restore a file from the trash]'
+	'--delete[Delete a file from the trash]'
+	'--empty[Empty the trash bin]'
+	'--rm[Escape hatch to permanently delete a file]'
+	'(-h --help)'{-h,--help}'[This help information]'
```
syohex, line 31:
```suggestion
	'(- *)'{-h,--help}'[This help information]'
```

Ditto

src/_trash
```diff
@@ -0,0 +1,49 @@
...
+	'(-r --recursive)'{-r,--recursive}'[Delete directories and their contents]'
+	'(-v --verbose)'{-v,--verbose}'[Print more information]'
+	'(-i --interactive)'{-i,--interactive}'[Ask before each deletion]'
+	'(-I --interactive-once)'{-I,--interactive-once}'[Ask once if deleting 3 or more]'
+	'(-f --force)'{-f,--force}'[Don'\''t prompt and ignore errors]'
+	'--version[Output the version and exit]'
+	'--list[List out the files in the trash]'
+	'--orphans[List orphaned files in the trash]'
+	'--restore[Restore a file from the trash]'
+	'--delete[Delete a file from the trash]'
```
syohex, line 28:
```suggestion
	'(--list --orphans)--list[List out the files in the trash]'
	'(--list --orphans)--orphans[List orphaned files in the trash]'
	'(--restore --delete)--restore[Restore a file from the trash]'
	'(--restore --delete)--delete[Delete a file from the trash]'
```

src/_trash
mirsella: nice didn't know we could do that

mirsella:
thank you for taking the time to improve the completion file, and comment your changes.

I think you deserve to add yourself to the authors list if you want :)

syohex:
Unfortunately I noticed that zsh itself already provides [another `trash` command completion](https://github.com/zsh-users/zsh/blob/master/Completion/Darwin/Command/_trash) and it is used by default with default setting instead of this completion file. (It is put under `Darwin` directory, but some Linux distributions set it to fpath)

I'm afraid that I think it is difficult to merge this completion file because
- This PR's completion file doesn't work without some specific configuration like modifying `fpath` variable etc
- It might break configurations for some users who use another trash command implementation

So I think you should write `trash` completion that works for both [macOS implementation](https://github.com/ali-rantakari/trash) and [D language implementation](https://github.com/rushsteve1/trash-d) and send it to [zsh mailing list](https://zsh.sourceforge.io/Arc/mlist.html) if you want to use `trash` command completion.

mirsella:
Thanks

I opened https://github.com/zsh-users/zsh/pull/99, if it doesn't succeed I'll try to reopen the PR on trash-d

PR #1035, opened 2023-08-31
Create _ngrok

omamkaz:


syohex:
I'm afraid that we cannot accept this PR by our policy(as below). `ngrok` supports zsh completion by itself(`ngrok completion` and this file content is equal to its output)

- Completions are not accepted when already available in their original project.

PR #1039, opened 2023-09-04
Add completion script for chatblade

johannes87:

### DIFF
--- a/src/_chatblade
+++ b/src/_chatblade
@@ -1,0 +1,57 @@
+#compdef chatblade
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for chatblade (https://github.com/npiv/chatblade)
+#
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * Johannes Bittner (https://github.com/johannes87) 
+#
+# ------------------------------------------------------------------------------
+
+_chatblade() {
+  local -a args
+  
+  args+=(
+    '(-h --help)'{-h,--help}'[show this help message and exit]'
+    '--openai-api-key[the OpenAI API key can also be set as env variable OPENAI_API_KEY]'
+    '--temperature[temperature (openai setting)]'
+    '(-c --chat-gpt)'{-c,--chat-gpt}'[chat GPT model 3.5/4 shorthand or full qualified model name, can also be set via e
+  variable OPENAI_API_MODEL]'
+    '(-i --interactive)'{-i,--interactive}'[start an interactive chat session. This will implicitly continue the
+  conversation]'
+    '(-s --stream)'{-s,--stream}'[Stream the incoming text to the terminal]'
+    '(-t --tokens)'{-t,--tokens}'[display what *would* be sent, how many tokens, and estimated costs]'
+    '(-p --prompt-file)'{-p,--prompt-file}'[prompt name - will load the prompt with that name at ~/.config/chatblade/nam
+  or a path to a file]:prompt file:_files'
+    '(-e --extract)'{-e,--extract}'[extract content from response if possible (either json or code block)]'
+    '(-r --raw)'{-r,--raw}'[print session as pure text, dont pretty print or format]'
+    '(-n --no-format)'{-n,--no-format}'[do not add pretty print formatting to output]'
+    '(-o --only)'{-o,--only}'[Only display the response, omit query]'
+    '--theme[Set the theme for syntax highlighting see https://pygments.org/styles/, can also be set with
+  CHATBLADE_THEME]:theme:_values "theme" default emacs friendly manni monokai'
+    '(-l --last)'{-l,--last}'[alias for -S last, the default session if none is specified]'
+    '(-S --session)'{-S,--session}'[initiate or continue named session]:session:->sessions'
+    '--session-list[list sessions]'
+    '--session-path[show path to session file]'
+    '--session-dump[dump session to stdout]'
+    '--session-delete[delete session]'
+    '--session-rename[rename session]:new session:->sessions'
+  )
+  
+  case $state in
+    (sessions)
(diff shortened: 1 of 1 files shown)


src/_chatblade
```diff
@@ -0,0 +1,57 @@
...
+    '(-s --stream)'{-s,--stream}'[Stream the incoming text to the terminal]'
+    '(-t --tokens)'{-t,--tokens}'[display what *would* be sent, how many tokens, and estimated costs]'
+    '(-p --prompt-file)'{-p,--prompt-file}'[prompt name - will load the prompt with that name at ~/.config/chatblade/nam
+  or a path to a file]:prompt file:_files'
+    '(-e --extract)'{-e,--extract}'[extract content from response if possible (either json or code block)]'
+    '(-r --raw)'{-r,--raw}'[print session as pure text, dont pretty print or format]'
+    '(-n --no-format)'{-n,--no-format}'[do not add pretty print formatting to output]'
+    '(-o --only)'{-o,--only}'[Only display the response, omit query]'
+    '--theme[Set the theme for syntax highlighting see https://pygments.org/styles/, can also be set with
+  CHATBLADE_THEME]:theme:_values "theme" default emacs friendly manni monokai'
```
syohex, line 36:
```suggestion
    '--theme[Set the theme for syntax highlighting see https://pygments.org/styles/, can also be set with CHATBLADE_THEME]:theme:_values "theme" default emacs friendly manni monokai'
```

Ditto

src/_chatblade
```diff
@@ -0,0 +1,57 @@
...
+#
+# ------------------------------------------------------------------------------
+
+_chatblade() {
+  local -a args
+  
+  args+=(
+    '(-h --help)'{-h,--help}'[show this help message and exit]'
+    '--openai-api-key[the OpenAI API key can also be set as env variable OPENAI_API_KEY]'
+    '--temperature[temperature (openai setting)]'
```
syohex, line 22:
```suggestion
    '--openai-api-key[the OpenAI API key can also be set as env variable OPENAI_API_KEY]:key'
    '--temperature[temperature (openai setting)]:temperature'
```

It seems these options take an argument

src/_chatblade
```diff
@@ -0,0 +1,57 @@
...
+#
+#  * Johannes Bittner (https://github.com/johannes87) 
+#
+# ------------------------------------------------------------------------------
+
+_chatblade() {
+  local -a args
+  
+  args+=(
+    '(-h --help)'{-h,--help}'[show this help message and exit]'
```
syohex, line 20:
```suggestion
    '(- *)'{-h,--help}'[show this help message and exit]'
```

If `--help` is specified, other options should not be completed.

src/_chatblade
```diff
@@ -0,0 +1,57 @@
...
+      local -a sessions
+      sessions=("${(@f)$(chatblade --session-list)}")
+      _describe 'session' sessions
+      ;;
+  esac
+  
+  _arguments -s $args
+}
+
+_chatblade
```
syohex, line 57:
```suggestion
_chatblade

# Local Variables:
# mode: Shell-Script
# sh-indentation: 2
# indent-tabs-mode: nil
# sh-basic-offset: 2
# End:
# vim: ft=zsh sw=2 ts=2 et
```

johannes87:
Thank you @syohex for your suggestions!

src/_chatblade
```diff
@@ -0,0 +1,61 @@
...
+# ------------------------------------------------------------------------------
+# Authors
+# -------
+#
+#  * Johannes Bittner (https://github.com/johannes87) 
+#
+# ------------------------------------------------------------------------------
+
+_chatblade() {
+  local -a args
```
syohex, line 17:
```suggestion
  local -a args
  local context state state_descr line
  typeset -A opt_args
```

Could you add those variable declarations to prevent global variables from overwriting ?

Reference
- https://zsh.sourceforge.io/Doc/Release/Completion-System.html

> A function calling _arguments with at least one action containing a ‘->string’ must therefore declare appropriate local parameters:

```
local context state state_descr line
typeset -A opt_args
```

src/_chatblade
```diff
@@ -0,0 +1,61 @@
...
+  
+  case $state in
+    (sessions)
+      local -a sessions
+      sessions=("${(@f)$(chatblade --session-list)}")
+      _describe 'session' sessions
+      ;;
+  esac
+  
+  _arguments -s $args
```
syohex, line 50:
```suggestion
  _arguments -s $args

  case $state in
    (sessions)
      local -a sessions
      sessions=("${(@f)$(chatblade --session-list)}")
      _describe 'session' sessions
      ;;
  esac
```

I tried `--session` argument completion but it didn't work. I think `_arguments` must be before `$state` case statement,  this works.

syohex (commented):
I'm afraid I still have a couple of comments please check them.

src/_chatblade
johannes87: I tried this change, but observed that the `--session` argument does no longer work after your suggestion. After changing it back, it works again.

johannes87: It seems to me that the "state" in the "local" builtin breaks this. I removed it and it works again (but I'm not sure why?)

src/_chatblade
johannes87: Seems to be because of the "prevent global variables from overwriting" change, please see my comments in the other thread

src/_chatblade
johannes87: Nevermind, I forgot to restart zsh. I applied both of your changes and now it works. Thanks!

src/_chatblade
johannes87: My fault =)

syohex (approved):

syohex:
Thanks

PR #1054, opened 2023-10-20
Add jest completion

aoyama-val:

### DIFF
--- a/src/_jest
+++ b/src/_jest
@@ -1,0 +1,169 @@
+#compdef jest jest.js
+# ------------------------------------------------------------------------------
+# Copyright (c) 2023 Github zsh-users - https://github.com/zsh-users
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * Neither the name of the zsh-users nor the
+#       names of its contributors may be used to endorse or promote products
+#       derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL ZSH-USERS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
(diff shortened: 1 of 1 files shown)


src/_jest
```diff
@@ -0,0 +1,170 @@
...
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# ------------------------------------------------------------------------------
+# Description
+# -----------
+#
+#  Completion script for Node.js 29.7.0 (https://jestjs.io/)
```
syohex, line 31:
```suggestion
#  Completion script for Jest 29.7.0 (https://jestjs.io/)
```

src/_jest
```diff
@@ -0,0 +1,170 @@
...
+  --changedSince'[Runs tests related to the changes since the provided branch. If the current branch has diverged from the given branch, then only changes made locally will be tested. Behaves similarly to `--onlyChanged`. (string)]:string' \
+  --ci'[Whether to run Jest in continuous integration (CI) mode. This option is on by default in most popular CI environments. It will prevent snapshots from being written unless explicitly requested. (boolean)]' \
+  --clearCache'[Clears the configured Jest cache directory and then exits. Default directory can be found by calling jest --showConfig (boolean)]' \
+  --clearMocks'[Automatically clear mock calls, instances, contexts and results before every test. Equivalent to calling jest.clearAllMocks() before each test. (boolean)]' \
+  --collectCoverage'[Alias for --coverage. (boolean)]' \
+  --collectCoverageFrom'[A glob pattern relative to <rootDir> matching the files that coverage info needs to be collected from. (string)]:string' \
+  --color'[Forces test results output color highlighting (even if stdout is not a TTY). Set to false if you would like to have no colors. (boolean)]' \
+  --colors'[Alias for `--color`. (boolean)]' \
```
syohex, line 60:
```suggestion
 '(--color --colors)'{--color,--colors}'[Forces test results output color highlighting (even if stdout is not a TTY). Set to false if you would like to have no colors. (boolean)]' \
```

We should not specify both `--color` and `--colors` so they should be merged. Or remove the alias version

src/_jest
```diff
@@ -0,0 +1,170 @@
...
+  --cacheDirectory'[The directory where Jest should store its cached dependency information. (string)]:directories:_dirs' \
+  --changedFilesWithAncestor'[Runs tests related to the current changes and the changes made in the last commit. Behaves similarly to `--onlyChanged`. (boolean)]' \
+  --changedSince'[Runs tests related to the changes since the provided branch. If the current branch has diverged from the given branch, then only changes made locally will be tested. Behaves similarly to `--onlyChanged`. (string)]:string' \
+  --ci'[Whether to run Jest in continuous integration (CI) mode. This option is on by default in most popular CI environments. It will prevent snapshots from being written unless explicitly requested. (boolean)]' \
+  --clearCache'[Clears the configured Jest cache directory and then exits. Default directory can be found by calling jest --showConfig (boolean)]' \
+  --clearMocks'[Automatically clear mock calls, instances, contexts and results before every test. Equivalent to calling jest.clearAllMocks() before each test. (boolean)]' \
+  --collectCoverage'[Alias for --coverage. (boolean)]' \
```
syohex, line 57: Please see the comment about `--color` and `--colors`

syohex:
image

Users cannot see a description as above if it is too long. So I think it should be one sentence at most except special cases.In most cases, it is enough to use the first sentence as the flag description.  And I think type information is redundant

aoyama-val:
Thank you for your review. Fixed them. I made the descriptions shorter, but the longest one is still 146 characters long. I don't think that's a real harml, rather benefit for users who have large monitors. What do you think?

syohex:
I think it's ok. thanks

PR #1072, opened 2024-03-04
feat(issue-1071): adds zellij completion

### LINKED ISSUE #1071: zsh-completions for zellij cli

dvrd:
Resolves #1071 

# Reasoning
Currently zellij doesn't automatically add completions, you have to add them yourself from a cli command provided by the project but is your decision how it gets sourced

## Changes
* Adds completion for zellij and utility functions

### DIFF
--- a/src/_zellij
+++ b/src/_zellij
@@ -1,0 +1,1134 @@
+#compdef zellij
+
+# ------------------------------------------------------------------------------
+# MIT License
+#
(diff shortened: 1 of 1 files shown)


syohex:

I'm afraid that we cannot accept its completion because zellij provides zsh completion by itself according to the above reason. We can generate it by `zellij setup --generate-completion zsh `

Issue #1075, opened 2024-04-07
Completions for fvm are broken as of fvm v3


bartekpacia:
Completion script in question: https://github.com/zsh-users/zsh-completions/blob/master/src/_fvm

As of latest `fvm` v3.0.0, the above completion script doesn't work. Learn more in https://github.com/leoafarias/fvm/issues/689

ping @syohex who created it (huge thank you and kudos!)

syohex:
Thanks for reporting issue. It seems fvm supports zsh completion itself now. I'll remove fvm completion from this repository according to our policy.

bartekpacia:
> It seems fvm supports zsh completion itself now. 

Hmm, does it? What do you mean?

syohex:
I tried new `fvm`. It generates completion files(e.g. `~/.dart-cli-completion/fvm.zsh`) under `~/.dart-cli-completion/` and it appends its configuration in the zsh configuration file(`.zshrc` etc)

bartekpacia:
>  It generates completion files(e.g. ~/.dart-cli-completion/fvm.zsh) under ~/.dart-cli-completion/ and it appends its configuration in the zsh configuration file(.zshrc etc)

Ah right, it uses [this package](https://pub.dev/packages/cli_completion). But it doesn't work on my machine for some reason. Thanks for checking!

> it appends its configuration in the zsh configuration file(.zshrc etc)

I really don't like this behavior. I want my completion scripts to live in e.g. `/opt/homebrew/share/zsh/site-functions/`, and not modify (!) my dotfiles and `~/.config`.

bartekpacia:
So yeah, let's remove `_fvm` from this repo.

Did you write that completion script manually?

syohex:
> Did you write that completion script manually?

Yes

syohex:
I've removed it at #1076

PR #1076, opened 2024-04-10
Remove _fvm completion

syohex:
Because _fvm supports zsh completion itself since version 3.0


syohex:
This is related to #1075

PR #1098, opened 2024-08-19
Add: Tsh completion implementation

AliKhalediPour:


syohex:
Thanks for PR but I'm afraid that we cannot accept this PR due to [our policy](https://github.com/zsh-users/zsh-completions/blob/master/CONTRIBUTING.md). It supports bash/zsh completion by itself

- https://github.com/gravitational/teleport/issues/3753
- https://github.com/gravitational/teleport/issues/30175

If you would like to improve the zsh completion, please send the PR to the original repo.

PR #1103, opened 2024-09-15
Add completions for sox

xyproto:

`sox` is a collection of executables and libraries that has been in Debian since 1997. It has also been in Arch Linux for at least 10 years (probably longer). It is available for most (all?) Linux distros.

I would argue that `sox` is more well known, and has more installations, than "the Play Framework for Java and Scala".

This PR proposes to replace `_play` (for "The Play Framework") with one for the `play` executable that comes with `sox`, and also add completions for `sox`, `soxi` and `rec`.

### DIFF
--- a/src/_rec
+++ b/src/_rec
@@ -1,9 +1,9 @@
-#compdef play
+#compdef rec
 # ------------------------------------------------------------------------------
 #
 # Description
 # -----------
-#  Completion script for the play command (from SoX)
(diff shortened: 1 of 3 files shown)


syohex:
I've merged sox_ng, play_nx, rec_ng completion at #1208.

syohex:
Yes it is defined as an alias of play_ng

PR #1104, opened 2024-09-15
Add 'go telemetry' completion

telemachus:


telemachus:
Go 1.23 added a `telemetry` command. This PR provides completion for `telemetry` itself and its three possible states (`off|local|on`).  I took the descriptions from the output of `go help telemetry`.

PR #1113, opened 2024-11-26
Update rspec completion

kyanagi:
* Fix broken `-I`
* Setting exclusive options
* Configuring options that can be used multiple times
* Add descriptions for each formatters


syohex:
Thanks

PR #1114, opened 2024-12-10
Add completion script for yfm (diplodoc cli)

egorlem:
Hi there 👋

Diplodoc - platform with open-source code for creating technical documentation based on the concept of Docs as Code. A simple and convenient document management solution for large and small teams.

Best regards

### DIFF
--- a/src/_yfm
+++ b/src/_yfm
@@ -74,26 +74,26 @@ _yfm() {
         publish)
           _arguments \
             '(--endpoint)'--endpoint'[Endpoint of S3 storage]:url:_urls' \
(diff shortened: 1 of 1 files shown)


syohex:
@egorlem I've updated your code according to the help document. Is this ok ? I tested it, it seems it works. Please let me know if you have any problems.

egorlem:
@syohex Thanks for the great work. The code looks and works amazing. I learned a lot about how to write a really good completion. I made a commit where I indicated your authorship. Is this appropriate?

syohex:
Thank you very much. It's my pleasure

PR #1125, opened 2025-03-03
Update TypeScript completion from 5.6 to 5.8

egorlem:

### 5.7

- Added targets `es2024` for the `--lib` and `--target` options.

### 5.8

- Added `--libReplacement` and `--erasableSyntaxOnly` options.
- Added `node18` target for the `--module` option. 

Best regards

### DIFF
--- a/src/_tsc
+++ b/src/_tsc
@@ -189,8 +189,8 @@ local -a opts=(
   '--synchronousWatchDirectory[Synchronously call callbacks and update the state of directory watchers on platforms]'
   '--excludeDirectories[Remove a list of directories from the watch process]:dirs:_files -/'
   "--excludeFiles[Remove a list of files from the watch mode's processing]:files:_files"
-  "--libReplacement[Enable library replacement]:boolean:(true false)"
-  "--erasableSyntaxOnly[Ensures that runtime constructs which are not part of ECMAScript are not allowed]:boolean:(true false)"
+  "--libReplacement[Enable library replacement]"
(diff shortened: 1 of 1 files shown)


egorlem:
erasableSyntaxOnly is a very cool thing. I've been waiting for this for a long time 
Screenshot 2025-03-03 at 16 50 36

src/_tsc
```diff
@@ -189,6 +189,8 @@ local -a opts=(
   '--synchronousWatchDirectory[Synchronously call callbacks and update the state of directory watchers on platforms]'
   '--excludeDirectories[Remove a list of directories from the watch process]:dirs:_files -/'
   "--excludeFiles[Remove a list of files from the watch mode's processing]:files:_files"
+  "--libReplacement[Enable library replacement]:boolean:(true false)"
+  "--erasableSyntaxOnly[Ensures that runtime constructs which are not part of ECMAScript are not allowed]:boolean:(true false)"
```
syohex, line 193:
```suggestion
  "--libReplacement[Enable library replacement]"
  "--erasableSyntaxOnly[Ensures that runtime constructs which are not part of ECMAScript are not allowed]"
```

They are boolean flags and we can omit its values so I suppose this looks better same as the other flags. I know they can accept `true` or `false` value however it is sometimes confusing for example `--erasableSyntaxOnly=true` causes syntax error, its flag parser doesn't work like other flag parsers do.

egorlem: This really seems redundant. Removed them

syohex:
Thanks

PR #1127, opened 2025-03-13
Update README.md - Zsh rebuilding the completion cache twice

SumathiD20:
Zsh rebuilding the completion cache twice

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,14 @@ Add `zinit light zsh-users/zsh-completions` to your `~/.zshrc`.
 
         rm -f ~/.zcompdump; compinit
 
+Second Solution:
+zsh-completions as a plugin, they suggest manually adding its source directory to fpath before sourcing oh-my-zsh.sh.
+This ensures Zsh loads the completion files only once, improving performance.
+Recommended Fix:
+Instead of the existing method, they propose adding this line:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,14 @@ Add `zinit light zsh-users/zsh-completions` to your `~/.zshrc`.
...
         rm -f ~/.zcompdump; compinit
 
+Second Solution:
+zsh-completions as a plugin, they suggest manually adding its source directory to fpath before sourcing oh-my-zsh.sh.
+This ensures Zsh loads the completion files only once, improving performance.
+Recommended Fix:
+Instead of the existing method, they propose adding this line:
+
+fpath=($ZSH/custom/plugins/zsh-completions/src $fpath)
+
```
ferdnyc, line 71: This information is already there. The advice for Oh-my-Zsh! was updated years ago in response to #603.

What you're proposing is exactly the same as:

> * Add it to `FPATH` in your `.zshrc` by adding the following line before `source "$ZSH/oh-my-zsh.sh"`:
>
>         fpath+=${ZSH_CUSTOM:-${ZSH:-~/.oh-my-zsh}/custom}/plugins/zsh-completions/src

...Which is exactly what the instructions _already say_.

ferdnyc (changes requested):
The addition is redundant, it offers a different wording of the same instructions that are already in the document.

PR #1260, opened 2026-04-23
fix yarn run completions

tborychowski:
`yarn run<TAB>` doesn't show scripts that have `:` in them (e.g. `lint:js`, `test:e2e`), which is a popular pattern. this change fixes that.


syohex:
Thanks

PR #1273, opened 2026-06-11
feat: add support for build-presets

arminveres:
When running `cmake --build --preset=*` the preset would not complete. 
Additionally fix `cmake --build --preset=* --target=*` target completion too, by looking for the first CMakeCache.txt in five levels of directories.

### DIFF
--- a/src/_cmake
+++ b/src/_cmake
@@ -144,6 +144,29 @@ _cmake_generator_options() {
   fi
 }
 
+# ----------------------
+# _cmake_build_presets
+# ----------------------
+(( $+functions[_cmake_build_presets] )) ||
+_cmake_build_presets() {
+  local list_presets=(${(f)"$(cmake --list-presets=build 2>/dev/null |
+    sed -n -e 's,^[[:space:]]*"\([^"]*\)"[[:space:]]*-[[:space:]]*\(.*\),\1:\2,p' \
+           -e 's,^[[:space:]]*"\([^"]*\)"[[:space:]]*$,\1,p')"})
+
+  _describe 'build presets' list_presets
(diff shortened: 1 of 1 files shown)


src/_cmake
```diff
@@ -144,6 +144,29 @@ _cmake_generator_options() {
...
+}
+
+# --------------------------
+# _cmake_preset_build_dir
+#
+# Finds the build directory by locating the nearest CMakeCache.txt.
+# --------------------------
+(( $+functions[_cmake_preset_build_dir] )) ||
+_cmake_preset_build_dir() {
+  find . -maxdepth 5 -name CMakeCache.txt -print -quit | xargs -I{} dirname {}
```
syohex: Please consider an error case, for example add `2>/dev/null`. I think `-quit` is not a portable option, some BSD find implementations don't implement this flag. `-maxdepth` is not a portable option too ([POSIX find](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html)) but it seems most implementations support it.

- https://man.netbsd.org/find.1
- https://man.openbsd.org/find.1

arminveres: Thanks @syohex for reviewing.
Good points, I did not consider non-portable options.
I will have to rework this anyhow, as -maxdepth does DFS and since some projects include submodule with possible CMakeCaches, it would at least be more heuristically correct to do BFS,
Further, if multiple presets are active, then only the first is used for target selection, which may not be correct.

arminveres: Hi @syohex Updated, removed `find` call and reverted to parsing the preset file, to properly complete on target, not just the first match.

PR #1278, opened 2026-07-02
add _xmake and _xrepo for xmake completion

DavidingPlus:

Added _xmake and _xrepo for xmake completion.


syohex:
I'm afraid that this is not acceptable according to this repo's policy.
See CONTRIBUTING.md.

PR #1293, opened 2026-08-16
Add yt-dlp completion

mmustafasenoglu:
This adds zsh completion for [yt-dlp](https://github.com/yt-dlp/yt-dlp), a popular video downloader with 100k+ stars.

yt-dlp does not ship its own zsh completion, so this fills that gap.

**What's covered:**
- All CLI options across 16 categories (general, network, geo-restriction, video selection, download, filesystem, thumbnail, internet shortcut, verbosity, workaround, format, subtitle, authentication, post-processing, sponsorblock, extractor)
- URL completion
- Type-specific completions (downloader names, browser names, format options, etc.)

**Checklist:**
- [x] This compdef is not already available in zsh
- [x] This compdef is not already available in the original project (yt-dlp has no zsh completion)
- [x] I am the original author


syohex:
> yt-dlp does not ship its own zsh completion, so this fills that gap.

It seems this is incorrect. `yt-dlp` provides its own completion file generated by [this script](https://github.com/yt-dlp/yt-dlp/blob/master/devscripts/zsh-completion.py). And many `yt-dlp` packages (Homebrew, Ubuntu/Debian, arch linux etc) bundle its zsh completion file.

I'm afraid that please send PR to official repository if you want to improve `yt-dlp` completion file.

